PR #3, opened 2020-09-02
Blob/Manifest validation retry

talal:


majewsky (approved):

majewsky:
Thanks. You don't need to take care of the deployment. I have another feature branch in the works and will take this one to prod when my feature is done.

PR #5, opened 2020-12-03
Secret handling

talal:


internal/keppel/config.go
```diff
@@ -139,3 +148,51 @@ func mayGetenvURL(key string) *url.URL {
...
+	}
+}
+
+// GetRedisURL returns a Redis connection URL by getting the required
+// parameters from environment variables:
+//   REDIS_PASSWORD, REDIS_HOSTNAME, REDIS_PORT, and REDIS_DB_NUM.
+//
+// The environment variable keys are prefixed with the provided prefix.
+func GetRedisURL(prefix string) string {
+	prefix = strings.Join([]string{prefix, "REDIS"}, "_")
```
majewsky, line 185:
```suggestion
	prefix = prefix + "_REDIS"
```

internal/keppel/config.go
```diff
@@ -139,3 +148,51 @@ func mayGetenvURL(key string) *url.URL {
...
+	dbNum := GetenvOrDefault(prefix+"_DB_NUM", "0")
+
+	url := url.URL{
+		Scheme: "redis",
+		User:   url.UserPassword("", pass),
+		Host:   net.JoinHostPort(host, port),
+		Path:   dbNum,
+	}
+	return url.String()
+}
```
majewsky, line 198: Instead of generating a URL that we then parse with `redis.ParseURL`, this should return a `*redis.Options` directly.

majewsky (changes requested):

majewsky:
I merged this manually because I did the `bump all dependencies` dance and only then noticed this PR was still around. I hope I didn't introduce any problems when merging the go.mod/go.sum

PR #8, opened 2021-05-28
Expose labels for image manifests

### LINKED ISSUE #7: expose labels in manifest list endpoint
Docker images can have labels, as defined by the `LABEL` command in a Dockerfile. For instance, in Keppel's own Dockerfile, we have:

https://github.com/sapcc/keppel/blob/3ea36f392d099d0380f26d478b22b92f3ccb10b9/Dockerfile#L10

Right now, if a user wants to retrieve those label values for a certain image, they have to download the manifest, then find the image configuration blob within the manifest, then download the image configuration blob, then parse it. That's a lot of work, and what's worse, several of these actions count towards rate limits, which makes surveying a large repo for image labels impractically slow. Keppel should provide image labels in its own API in a more convenient spot.

In fact, Keppel already looks at the image labels to enforce the `required_labels` option that can be enabled on the account level, so it should be fairly straight-forward to retain the label values in Keppel's DB for quick access. For simplicity's sake, it will be sufficient to store the labels as JSON in a text column on the manifest object.

- [x] add a text field `labels_json` to the `manifests` table, and extend the `keppel.Manifest` struct accordingly
- [x] fill this field during `func validateAndStoreManifestCommon` (the existing testcase for the `required_labels` option can be used to validate that the field is populated correctly)
- [x] add a `Labels` field to struct `keppelv1.Manifest` (probably using the `json.RawMessage` type) and fill it in the handler method for `GET /keppel/v1/repositories/:repo/manifests`
- [x] adjust the unit test for the API, as well as the API docs, accordingly

No extra migration should be needed for existing images. `validateAndStoreManifestCommon` is called periodically on existing manifests by the janitor, so the new field should be slowly populated once the changes are rolled out.

talal:
Fixes #7


internal/api/keppel/manifests.go
```diff
@@ -102,12 +104,14 @@ func (a *API) handleGetManifests(w http.ResponseWriter, r *http.Request) {
 			result.IsTruncated = true
 			break
 		}
+		labelsJSON := []byte(dbManifest.LabelsJSON)
```
majewsky, line 107: This tripped me up when reading since there are two casts happening in succession here. This could use some clarification:
```suggestion
		labelsJSON := json.RawMessage([]byte(dbManifest.LabelsJSON))
```

internal/api/registry/manifests_test.go
```diff
@@ -573,6 +574,31 @@ func TestManifestRequiredLabels(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err.Error())
+		}
+		expected := map[string]string{
+			"bar": "is there",
+			"foo": "is there",
+		}
+		if !assert.DeepEqual(t, "labels_json", actual, expected) {
+			t.Fatal("expected labels_json to be correctly populated during image push")
+		}
```
majewsky, line 601: `assert.DeepEqual()` already prints an error message if the assertion fails, so the additional custom message is not required.
```suggestion
		assert.DeepEqual(t, "labels_json", actual, expected)
```

majewsky (approved):
Very nice. Two small nitpicks, then you can merge and deploy to QA right away:

PR #12, opened 2021-09-17
Delete image lists before manifests in account deletion

SuperSandro2000:
to remove all contents of accounts without triggering foreign key constraints

### DIFF
--- a/internal/api/keppel/accounts_test.go
+++ b/internal/api/keppel/accounts_test.go
@@ -38,7 +38,7 @@ import (
 	"github.com/sapcc/keppel/internal/test"
 )
 
-func setup(t *testing.T) (http.Handler, *test.AuthDriver, *test.FederationDriver, *test.Auditor, keppel.StorageDriver, *keppel.DB, *test.ClairDouble) {
+func setup(t *testing.T) (http.Handler, *test.AuthDriver, *test.FederationDriver, *test.Auditor, keppel.StorageDriver, *keppel.DB, *test.Clock, *test.ClairDouble) {
 	cfg, db := test.Setup(t, nil)
 
 	//setup a dummy ClairClient for testing the GET vulnerability report endpoint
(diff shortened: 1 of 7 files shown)


internal/api/keppel/accounts_test.go
```diff
@@ -1685,6 +1715,15 @@ func TestDeleteAccount(t *testing.T) {
...
+	assert.HTTPRequest{
+		Method: "DELETE",
+		Path: fmt.Sprintf(
+			"/keppel/v1/accounts/test1/repositories/%s/_manifests/%s",
+			repos[0].Name, "sha256:13493afd2a64d1a14fb5ca0d2fe312117239b935ea40dc35ca59c55931a86ab7",
+		),
+		Header:       map[string]string{"X-Test-Perms": "view:tenant1,delete:tenant1"},
+		ExpectStatus: http.StatusNoContent,
+	}.Check(t, r)
 	assert.HTTPRequest{
```
majewsky, line 1727: I would like to see another test on `DELETE /keppel/v1/accounts/test1` to see that it now asks for the non-list manifest to be deleted.

internal/api/keppel/accounts_test.go
```diff
@@ -1633,6 +1656,13 @@ func TestDeleteAccount(t *testing.T) {
 		}
 	}
 
+	imageList := test.GenerateImageList(image.Manifest)
```
majewsky, line 1659: I changed the type signature of the GenerateImageList function [yesterday](https://github.com/sapcc/keppel/commit/6dd6c27a03b377b8bdafbe6ccb8b625ff462807c), so this needs to be adapted:
```suggestion
	imageList := test.GenerateImageList(image)
```

majewsky (changes requested):
Thanks, looks great already!

majewsky (approved):
Bjutiful.

PR #15, opened 2021-09-21
Add endpoint to peer api to query an account's platform filter

SuperSandro2000:


internal/api/keppel/accounts.go
```diff
@@ -542,6 +544,66 @@ func (a *API) handlePutAccount(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
+		// Copy PlatformFilter when creating an account with the Replication Policy on_first_use
+		if req.Account.ReplicationPolicy != nil {
```
majewsky, line 548: This branch should only be taken when an account is being created, i.e. `if account == nil` (in other words, it should go in the `if account == nil` that ended just above these lines).

internal/api/keppel/accounts.go
```diff
@@ -542,6 +544,66 @@ func (a *API) handlePutAccount(w http.ResponseWriter, r *http.Request) {
...
+				if respondwith.ErrorText(w, err) {
+					return
+				}
+
+				reqURL := fmt.Sprintf("https://%s/peer/v1/account-filter/%s", accountToCreate.UpstreamPeerHostName, accountToCreate.Name)
+				authReq, err := http.NewRequest("GET", reqURL, nil)
+				if respondwith.ErrorText(w, err) {
+					return
+				}
+				//TODO: a.cfg.OurReplicationUsername = fmt.Sprintf("replication@%s", a.cfg.APIPublicURL.Hostname())
```
majewsky, line 566: There's some left-over code here.

internal/api/keppel/accounts.go
```diff
@@ -542,6 +544,66 @@ func (a *API) handlePutAccount(w http.ResponseWriter, r *http.Request) {
...
+
+				resp, err := http.DefaultClient.Do(authReq)
+				if err != nil {
+					http.Error(w, "could not validate credentials: "+err.Error(), http.StatusUnauthorized)
+					return
+				}
+				if resp.StatusCode != http.StatusOK {
+					http.Error(w, "could not validate credentials: expected 200 OK, but got "+resp.Status, http.StatusUnauthorized)
+					return
+				}
```
majewsky, line 577: "could not validate credentials" does not match with what you're doing here. How about "could not validate platform filter"?

internal/api/keppel/accounts.go
```diff
@@ -542,6 +544,66 @@ func (a *API) handlePutAccount(w http.ResponseWriter, r *http.Request) {
...
+
+				respBodyBytes, err := ioutil.ReadAll(resp.Body)
+				if err == nil {
+					err = resp.Body.Close()
+				} else {
+					resp.Body.Close()
+				}
+				if err != nil {
+					return
+				}
```
majewsky, line 587: Each branch that does `return` should have rendered a HTTP response beforehand:
```suggestion
				if respondwith.ErrorText(w, err) {
					return
				}
```

internal/api/keppel/accounts.go
```diff
@@ -542,6 +544,66 @@ func (a *API) handlePutAccount(w http.ResponseWriter, r *http.Request) {
...
+
+				var data peerv1.AccountFilter
+				err = json.Unmarshal(respBodyBytes, &data)
+				if err != nil {
+					return
+				}
```
majewsky, line 593: Same as above.
```suggestion
				if respondwith.ErrorText(w, err) {
					return
				}
```

internal/api/keppel/accounts.go
```diff
@@ -542,6 +544,66 @@ func (a *API) handlePutAccount(w http.ResponseWriter, r *http.Request) {
...
+
+				if req.Account.PlatformFilter == nil {
+					accountToCreate.PlatformFilter = data.Filter
+				} else if !reflect.DeepEqual(req.Account.PlatformFilter, data.Filter) {
+					// check if the peer PlatformFilter matches the primary account PlatformFilter
+					JSONPlatformFilter, _ := json.Marshal(req.Account.PlatformFilter)
+					JSONFilter, _ := json.Marshal(data.Filter)
+					http.Error(w, fmt.Sprintf("peer account filter needs to match primary account filter: primary account %s, peer account %s ", JSONPlatformFilter, JSONFilter), http.StatusConflict)
```
majewsky, line 601: Identifiers for locally-scoped variables should start in lower case by convention. (I'm actually surprised that staticcheck does not flag this.)
```suggestion
					jsonPlatformFilter, _ := json.Marshal(req.Account.PlatformFilter)
					jsonFilter, _ := json.Marshal(data.Filter)
					http.Error(w, fmt.Sprintf("peer account filter needs to match primary account filter: primary account %s, peer account %s ", jsonPlatformFilter, jsonFilter), http.StatusConflict)
```

Also, line 601 is quite long and could be broken up a bit. When messages get very long, I usually send the sprintf into a temp variable first, like so:
```go
msg := fmt.Sprintf("peer account filter needs to match primary account filter: primary account %s, peer account %s ",
  jsonPlatformFilter, jsonFilter)
http.Error(w, msg, http.StatusConflict)
```

majewsky (changes requested):

majewsky (approved):

PR #20, opened 2021-10-01
Support ManifestReference with both tag and digest

SuperSandro2000:


internal/keppel/imageref_test.go
```diff
@@ -74,3 +74,22 @@ func TestParseImageReferenceSuccess(t *testing.T) {
 		}
 	}
 }
+
+func TestParseImageReferenceLabelDigestSuccess(t *testing.T) {
+	registry := "localhost:5000"
+	repo := "library/alpine"
+	digest := "sha256:e9707504ad0d4c119036b6d41ace4a33596139d3feb9ccb6617813ce48c3eeef"
+	// Check that the image reference :nonsense@digest is equal to @digest
```
majewsky, line 82: Please make this comment describe more explicitly that we're maintaining bug-for-bug compatibility with the Docker client here.

majewsky (approved):
Can be merged when the following is addressed:

majewsky (approved):

PR #21, opened 2021-10-04
Add prometheus alert to check anycast membership

SuperSandro2000:


cmd/anycastmonitor/main.go
```diff
@@ -135,3 +153,40 @@ func (j *anycastMonitorJob) ValidateImages(manifestRef keppel.ManifestReference)
 		}
 	}
 }
+
+func (j *anycastMonitorJob) ValidateAnycastMembership(apiPublicHostname string) {
+	resp, err := http.Get("https://keppel.global.cloud.sap/keppel/v1/auth?service=keppel.global.cloud.sap&scope=repository:foo/bar:pull")
```
majewsky, line 158: Our hostnames should not be hardcoded here.

cmd/anycastmonitor/main.go
```diff
@@ -135,3 +153,40 @@ func (j *anycastMonitorJob) ValidateImages(manifestRef keppel.ManifestReference)
...
+	if err != nil {
+		logg.Error("failed reading body: %s", err.Error())
+	}
+
+	var data authapi.TokenResponse
+	err = json.Unmarshal(body, &data)
+	if err != nil {
+		logg.Error("failed to unmarshal JWT: %s", err.Error())
+	}
+	token, err := base64.StdEncoding.DecodeString(strings.Split(data.Token, ".")[1])
```
majewsky, line 172: You should check the length of the slice returned from `strings.Split` before proceeding, to avoid a crash if `data.Token` is malformed and contains no dots at all.

majewsky (changes requested):

majewsky (approved):
Please go ahead and merge together with the respective Helm chart change.

PR #24, opened 2021-11-18
add support for ed25519 keys and seamless key rotation

majewsky:
ed25519 support improves token size and speeds up token issuance as well as token validation, but it requires a slight amount of refactoring to get rid of libtrust, which is unmaintained and never received support for ed25519 keys. Thankfully, we can get all the required functionality from std and our JWT library.

To enable a seamless transition to ed25519 keys, I'm also adding support for key rotation, where old keys can still be used to validate tokens, but new tokens will be issued with the most recent key only.


internal/auth/token.go
```diff
@@ -129,36 +140,50 @@ func (a Authorization) IssueToken(cfg keppel.Configuration) (*TokenResponse, err
...
+}
+
+func serializePublicKey(key crypto.PrivateKey) string {
+	switch key := key.(type) {
+	case ed25519.PrivateKey:
+		pubkey := key.Public().(ed25519.PublicKey)
+		return hex.EncodeToString([]byte(pubkey))
+	case *rsa.PrivateKey:
+		pubkey := key.Public().(*rsa.PublicKey)
+		return fmt.Sprintf("%x:%s", pubkey.E, pubkey.N.Text(16))
```
SuperSandro2000, line 184:
```suggestion
		return fmt.Sprintf("%x:%x", pubkey.E, pubkey.N)
```
Why are we not doing this? Is there a catch here?

majewsky: This won't work because `pubkey.N` is a bigint. We need to use its custom methods for formatting.

PR #25, opened 2021-11-24
add domain-remapping middleware incl. documentation

majewsky:
@SuperSandro2000 I will write the tests tomorrow, but you can already have a look at the implementation part if you like.

### DIFF
--- a/docs/operator-guide.md
+++ b/docs/operator-guide.md
@@ -8,6 +8,7 @@ In this document:
   - [Drivers](#drivers)
   - [Common configuration options](#common-configuration-options)
   - [API server configuration options](#api-server-configuration-options)
+  - [API server: Domain remapping support](#api-server-domain-remapping-support)
   - [Janitor configuration options](#janitor-configuration-options)
 - [Prometheus metrics](#prometheus-metrics)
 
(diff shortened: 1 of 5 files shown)


docs/operator-guide.md
```diff
@@ -206,6 +207,17 @@ These options are only understood by the API server.
...
+### API server: Domain remapping support
+
+Usually, Keppel exposes its APIs under the hostnames specified in `$KEPPEL_API_PUBLIC_URL` and `$KEPPEL_API_ANYCAST_URL`. However, if you wish, you can also configure your HTTPS reverse-proxy to serve the Keppel API on direct subdomains of these hostnames. In this case, the name of the subdomain will be interpreted as a Keppel account name, and the Registry API will be exposed on these subdomains without requiring the account name in the URL path.
+
+For example, if `KEPPEL_API_PUBLIC_URL=https://registry.example.com`, then the image `latest` in the repository `myrepo` in the account `myaccount` can be reached via the following URLs:
+
+- standard URL: `https://registry.example.com/v2/myaccount/myrepo/manifests/latest`
+- domain-remapped URL: `https://myaccount.registry.example.com/v2/myrepo/manifests/latest`
+
+This is particularly useful if you have an external replica account of Docker Hub, since dockerd only accepts plain hostnames in its `registry-mirrors` option. Therefore something like `registry.example.com/dockerhubmirror` would not work, but `dockerhubmirror.registry.example.com` does work.
```
SuperSandro2000, line 219:
```suggestion
This is particularly useful if you have an external replica account of Docker Hub, since the docker daemon only accepts plain hostnames in its `registry-mirrors` option. Therefore something like `registry.example.com/dockerhubmirror` would not work, but `dockerhubmirror.registry.example.com` does work.
```

Should we provide an example json snippet which people can copy into their daemon.json files?

internal/api/domain_remap.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+//ServeHTTP implements the http.Handler interface.
+func (h domainRemap) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	//if request does not use domain remapping, forward to the next handler unchanged
+	rewrittenURL, matches := h.tryRewriteURL(keppel.OriginalRequestURL(r))
+	if !matches {
+		h.next.ServeHTTP(w, r)
+		return
+	}
```
SuperSandro2000, line 55: I think we should differentiate between no domain remapping and wrong domain remapping before people start to use their subdomains and we can't easily remove it anymore.  eg ``https://myaccount.registry.example.com/v2/myaccount/myrepo/manifests/latest`` should error differently and shouldn't be passed through.

docs/operator-guide.md
majewsky: That's an operator handbook. I don't think we need that much hand-holding.

internal/api/domain_remap.go
majewsky: Your example URL is legal though. Repos can have a path, so in your case the account would be `myaccount` and the repo would be `myaccount/myrepo`, corresponding to `https://registry.example.com/v2/myaccount/myaccount/myrepo/manifests/latest`. It's not unreasonable to assume that such scenarios would arise in real-life usage, e.g. if the account is named after a team's main product.

In fact, I did a small search for `SELECT * FROM repos WHERE name LIKE CONCAT(account_name, '/%')` and found several such repos in prod.

SuperSandro2000: and what would happen if ``myaccount.registry.example.com/myaccount/myrepo/manifests/latest`` was used?

majewsky: Good point, I've added some extra logic to explicitly forbid anything that's not `/v2/...`.

PR #27, opened 2021-12-07
Add match_cidr to rbac_policies

SuperSandro2000:
TODO:
- tests for cidr
- another test for rbac anon_first_pull


internal/keppel/database.go
```diff
@@ -439,6 +439,30 @@ var sqlMigrations = map[string]string{
...
+		ALTER TABLE rbac_policies
+			ADD COLUMN match_cidr TEXT NOT NULL DEFAULT '0.0.0.0/0';
+		ALTER TABLE rbac_policies
+			ADD PRIMARY KEY (account_name, match_cidr, match_repository, match_username);
+	`,
+	"027_add_rbac_policies_match_cidr.down.sql": `
+		ALTER TABLE rbac_policies
+			DROP COLUMN match_cidr;
+		ALTER TABLE rbac_policies
+			DROP CONSTRAINT rbac_policies_pkey;
```
majewsky, line 454: You need to drop the constraint before being able to drop one of the columns it refers to.

internal/keppel/database.go
```diff
@@ -439,6 +439,30 @@ var sqlMigrations = map[string]string{
...
+		ALTER TABLE rbac_policies
+			ADD PRIMARY KEY (account_name, match_repository, match_username);
+	`,
+	"028_add_rbac_policies_can_anonymous_first_pull .up.sql": `
+	ALTER TABLE rbac_policies
+		ADD COLUMN can_anonymous_first_pull BOOLEAN NOT NULL DEFAULT FALSE;
+	`,
+	"028_add_rbac_policies_can_anonymous_first_pull .down.sql": `
+	ALTER TABLE rbac_policies
+		DROP CONSTRAINT can_anonymous_first_pull ;
```
majewsky, line 464:
```suggestion
		DROP COLUMN can_anonymous_first_pull ;
```

internal/keppel/database.go
```diff
@@ -439,6 +439,30 @@ var sqlMigrations = map[string]string{
...
+	`,
+	"027_add_rbac_policies_match_cidr.down.sql": `
+		ALTER TABLE rbac_policies
+			DROP COLUMN match_cidr;
+		ALTER TABLE rbac_policies
+			DROP CONSTRAINT rbac_policies_pkey;
+		ALTER TABLE rbac_policies
+			ADD PRIMARY KEY (account_name, match_repository, match_username);
+	`,
+	"028_add_rbac_policies_can_anonymous_first_pull .up.sql": `
```
majewsky, line 458: Filename contains a space (same for the `.down.sql` below). It looks like the migrate lib accepts that, but I'd rather have it without space for future-proofing.

docs/api-spec.md
```diff
@@ -171,7 +171,7 @@ The following fields may be returned:
...
 | `accounts[].rbac_policies[].match_username` | string | The RBAC policy applies to all users whose name matches this regex. Refer to the [documentation of your auth driver](./drivers/) for the syntax of usernames. The notes on regexes below apply. |
-| `accounts[].rbac_policies[].permissions` | list of strings | The permissions granted by the RBAC policy. Acceptable values include `pull`, `push`, `delete` and `anonymous_pull`. When `pull`, `push` or `delete` are included, `match_username` is not empty. When `anonymous_pull` is included, `match_username` is empty. |
+| `accounts[].rbac_policies[].permissions` | list of strings | The permissions granted by the RBAC policy. Acceptable values include `pull`, `push`, `delete`, `anonymous_pull` and `first_pull`. When `pull`, `push` or `delete` are included, `match_username` is not empty. When `anonymous_pull` or `first_pull` is included, `match_username` is empty. |
```
majewsky, line 174: "first_pull" should be called "anonymous_first_pull" to clarify that it is relevant specifically for anonymous users. Also, this line should include a short explanation what this permission means (and that it is limited to external replica accounts), since it is not as obvious as the other ones.

internal/api/keppel/accounts.go
```diff
@@ -394,6 +415,10 @@ func (a *API) handlePutAccount(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, err.Error(), http.StatusUnprocessableEntity)
 			return
 		}
+		if rbacPolicies[idx].CanFirstPullAnonymously && req.Account.ReplicationPolicy == nil {
```
majewsky, line 418: This only checks for replication, not for external replication:
```suggestion
		rp := req.Account.ReplicationPolicy
		if rbacPolicies[idx].CanFirstPullAnonymously && (rp == nil || rp.ExternalPeer.URL == "") {
```

internal/api/keppel/accounts_test.go
```diff
@@ -1000,6 +1000,54 @@ func TestPutAccountErrorCases(t *testing.T) {
...
+				"auth_tenant_id": "tenant1",
+				"rbac_policies": []assert.JSONObject{{
+					"match_repository": "library/.+",
+					"permissions":      []string{"anonymous_first_pull"},
+				}},
+			},
+		},
+		ExpectStatus: http.StatusUnprocessableEntity,
+		ExpectBody:   assert.StringData("RBAC policy with \"can_anonymous_first_pull\" may only be for external replica accounts\n"),
+	}.Check(t, h)
```
majewsky, line 1050: There should be additional tests where:
- an RBAC policy with `anonymous_first_pull` is successfully set and subsequently successfully retrieved, and where
- an RBAC policy with `match_cidr` equal to `0.0.0.0/0` is set and the retrieval does not show this field.

internal/api/registry/manifests.go
```diff
@@ -65,8 +66,15 @@ func (a *API) handleGetOrHeadManifest(w http.ResponseWriter, r *http.Request) {
...
 			//when replicating from external, only authenticated users can trigger the replication
 			if account.ExternalPeerURL != "" && authz.UserIdentity.UserType() != keppel.RegularUser {
-				keppel.ErrDenied.With("image does not exist here, and anonymous users may not replicate images").WithStatus(http.StatusForbidden).WriteAsRegistryV2ResponseTo(w, r)
-				return
+				// TODO: make this check less brittle by using contains
+				if !authz.ScopeSet.Contains(auth.Scope{
+					ResourceType: "repository",
+					ResourceName: repo.FullName(),
+					Actions:      []string{"pull", "first_pull"},
+				}) {
```
majewsky, line 74: It looks like this TODO was already checked off the list?

internal/api/registry/replication_test.go
```diff
@@ -360,7 +371,64 @@ func TestReplicationForbidAnonymousReplicationFromExternal(t *testing.T) {
...
+				Token string `json:"token"`
+			}
+			err = json.Unmarshal(tokenBodyBytes, &tokenBodyData)
+			if err != nil {
+				t.Fatal(err.Error())
+			}
+			anonToken := tokenBodyData.Token
 
+			// the rbac policy allows to replicate test1/foo images
+			expectManifestExists(t, h2, anonToken, "test1/foo", image.Manifest, "first", nil)
```
majewsky, line 431: Because of the `expectManifestExists` with an authenticated token earlier on, the image is already replicated, therefore the anonymous_first_pull permission is irrelevant. You should skip that first `expectManifestExists`.

internal/api/registry/replication_test.go
```diff
@@ -322,10 +322,21 @@ func TestReplicationForbidAnonymousReplicationFromExternal(t *testing.T) {
...
+			// enable anonymous pull and replication on test2/bar
+			err = s2.DB.Insert(&keppel.RBACPolicy{
+				AccountName:             "test1",
+				RepositoryPattern:       "bar",
+				CanPullAnonymously:      true,
+				CanFirstPullAnonymously: true,
+			})
+			if err != nil {
+				t.Fatal(err.Error())
+			}
```
majewsky, line 334: Why are we setting an RBAC policy on an irrelevant repository here?

internal/keppel/database.go
```diff
@@ -439,6 +439,30 @@ var sqlMigrations = map[string]string{
...
+	ALTER TABLE rbac_policies
+		ADD COLUMN can_anonymous_first_pull BOOLEAN NOT NULL DEFAULT FALSE;
```
majewsky, line 460: Since we already have a column `can_anon_pull`, it would be more consistent to call the new column `can_anon_first_pull`.

internal/api/keppel/accounts.go
```diff
@@ -227,10 +236,22 @@ func parseRBACPolicy(policy RBACPolicy) (keppel.RBACPolicy, error) {
...
 	}
+	// validate cidr early to prevent errors
+	// this has also the nice side effect that we can use the cidr of the network incase an ip is used
+	if cidr := policy.CidrPattern; cidr != "" {
+		_, net, err := net.ParseCIDR(cidr)
+		if err != nil {
+			// err.Error() sadly does not contain any useful information why the cidr is invalid
+			return result, fmt.Errorf("%q is not a valid cidr", cidr)
+		}
+		result.CidrPattern = net.String()
```
majewsky, line 247: When rendering existing policies, we treat `0.0.0.0/0` as "unset", but when taking in new policies, we don't apply the same rigor. I think the minimal required change is:

```suggestion
		result.CidrPattern = net.String()
		if result.CidrPattern == "0.0.0.0/0" {
			result.CidrPattern = ""
		}
```

Also, some test coverage would be useful. At least, having a policy with only `match_cidr = "0.0.0.0/0"` and some permissions should cause the complaint that no matches are set at all.

majewsky (changes requested):
Do not find yourself discouraged by the amount of notes, most of it is minor stuff. I'm mostly impressed by how well you find your way around Go and Keppel, given all the weird idiosyncrasies that you are rightfully pointing out.

internal/api/registry/replication_test.go
SuperSandro2000: On second thought this should be tested in the policy_test.

majewsky:
I will merge this now in order to start working on the UI changes. Please provide the remaining testcases in a followup PR.

PR #30, opened 2021-12-16
Manifest layer age

SuperSandro2000:


Makefile.maker.yaml
```diff
@@ -33,3 +33,6 @@ verbatim: |
   # This is for manual testing.
   run-api: build/keppel
     set -euo pipefail && source ./.env && $(CURDIR)/build/keppel server api
+
+  copy-fixtures:
+    fd '.actual' | xargs -I{} bash -c 'mv {} $$(echo {} | sed "s/.actual//g")'
```
majewsky, line 38: fd is a non-standard utility and can easily be replaced by a standard one:
```suggestion
    find -name '*.actual' | xargs -I{} bash -c 'mv {} $$(echo {} | sed "s/.actual//g")'
```

internal/api/registry/manifests_test.go
```diff
@@ -425,8 +425,11 @@ func TestImageListManifestLifecycle(t *testing.T) {
 		deleteToken := s.GetToken(t, "repository:test1/foo:delete")
 
 		//as a setup, upload two images and render a third image that's not uploaded
+		//advance time between image generation to test {Min,Max}LayerCreatedAt
 		image1 := test.GenerateImage(test.GenerateExampleLayer(1))
+		s.Clock.Step()
 		image2 := test.GenerateImage(test.GenerateExampleLayer(2))
+		s.Clock.Step()
 		image3 := test.GenerateImage(test.GenerateExampleLayer(3))
```
majewsky, line 433: The clock steps are unnecessary since the generator functions do not inspect `s.Clock`.

majewsky (changes requested):

majewsky (approved):
Looks good. Please hold off the merge until beginning of next year. I want to keep `master` clean across all repos just in case someone needs to make an emergency fix.

PR #32, opened 2022-02-10
Required labels lists

SuperSandro2000:


internal/api/registry/manifests_test.go
```diff
@@ -662,3 +662,91 @@ func TestImageManifestWrongBlobSize(t *testing.T) {
...
+	testWithPrimary(t, nil, func(s test.Setup) {
+		h := s.Handler
+		token := s.GetToken(t, "repository:test1/foo:pull,push")
+
+		baseImageConfig := test.BaseImageConfig
+		config, ok := baseImageConfig["config"].(map[string]interface{})
+		if ok {
+			config["Labels"] = map[string]string{"foo": "is there", "bar": "is there"}
+		}
+		baseImageConfig["config"] = config
```
majewsky, line 676: This breaks other tests, since you're editing the insides of `test.BaseImageConfig` that other tests will then use. (I'm guessing that this is why you ended up with all these fixture changes from unrelated tests.)

I suggest you convert `baseImageConfig` into a `func BaseImageConfig() map[string]interface{}` that returns a fresh config instance for each time it is called.

internal/api/registry/manifests_test.go
```diff
@@ -662,3 +662,91 @@ func TestImageManifestWrongBlobSize(t *testing.T) {
...
+		}.Check(t, h)
+
+		// image list manifest push with labels should succeed
+		list2 := test.GenerateImageList(image2)
+		assert.HTTPRequest{
+			Method: "PUT",
+			Path:   "/v2/test1/foo/manifests/latest",
+			Header: map[string]string{
+				"Authorization": "Bearer " + token,
+				"Content-Type":  "application/vnd.docker.distribution.manifest.v2+json",
```
majewsky, line 722: You need a different MIME type here:
```suggestion
				"Content-Type":  "application/vnd.docker.distribution.manifest.list.v2+json",
```

majewsky (changes requested):

majewsky (approved):

PR #36, opened 2022-03-18
Switch to golangci-lint

talal:


internal/api/registry/uploads.go
```diff
@@ -417,8 +418,10 @@ func (a *API) handleFinishBlobUpload(w http.ResponseWriter, r *http.Request) {
 		contentLength, err := strconv.ParseUint(contentLengthStr, 10, 64)
 		if err != nil {
 			//COVERAGE: unreachable in unit tests because net/http validates Content-Length header format before sending
-			keppel.ErrSizeInvalid.With("malformed Content-Length: " + err.Error())
-			return
+			err = keppel.ErrSizeInvalid.With("malformed Content-Length: " + err.Error())
+			if respondWithError(w, r, err) {
+				return
+			}
```
majewsky, line 424: Nice catch from the lint. But the common pattern to report errors of concrete types is
```suggestion
			keppel.ErrSizeInvalid.With("malformed Content-Length: " + err.Error()).WriteAsRegistryV2ResponseTo(w, r)
			return
```

internal/api/registry/blobs_test.go
```diff
@@ -382,7 +381,7 @@ func TestBlobStreamedAndChunkedUpload(t *testing.T) {
 					Body:         assert.ByteData(blob.Contents),
 					ExpectStatus: http.StatusAccepted,
 				}.Check(t, h)
-				uploadURL := resp.Header.Get("Location")
+				uploadURL := resp.Header.Get("Location") //nolint:govet
```
majewsky, line 384: Just out of curiosity, what was govet complaining about here?

majewsky (approved):
Thanks! One nitpick and one question, then this can go in.

internal/api/registry/blobs_test.go
talal: Variable shadowing of `uploadURL`.

majewsky: I see, thanks!

PR #37, opened 2022-03-24
Switch to gofrs/uuid library

talal:
See https://pkg.go.dev/vuln/GO-2020-0018.

We already use `gofrs/uuid` elsewhere in our other services.


majewsky:
Thanks! I will take that along to prod with my own changes on Monday.

PR #38, opened 2022-03-29
Report manifests as unsupported when they are to big

SuperSandro2000:


majewsky (changes requested):
I've updated `MergeVulnerabilityStatuses` to work well with the new value, but there are two big concerns about the task implementation for me:

1. The task still forces replication of the big images even though we don't end up scanning them. It should never call `Processor.ReplicateBlob` for a big image.
2. I don't see any assert in the test verify that the big image actually has the correct status and scan error set on it.

SuperSandro2000:
> The task still forces replication of the big images even though we don't end up scanning them. It should never call `Processor.ReplicateBlob` for a big image.

moved the code before the replication.

> I don't see any assert in the test verify that the big image actually has the correct status and scan error set on it.

That was an error in my test code. I only changed the size of the manifest and summed the size over the blobs. I now take the size of the manifest which should be correct because IIRC we check that in the consistency checks once in a while.

majewsky (approved):

PR #39, opened 2022-04-04
ignore values equal to Unix epoch when aggregating layer creation timestamps

majewsky:
There are tons of images reporting `X-Keppel-Min-Layer-Created-At: 0` since they are based on `FROM gcr.io/distroless` which is built reproducibly and thus does not set the build timestamp correctly: <https://github.com/GoogleContainerTools/distroless/issues/112>

This change causes Keppel to ignore such layers and only use timestamps that are not equal to the Unix epoch. The unit tests had to be reworked a bit since internal/api/registry had no tests for non-epoch creation timestamps yet.


SuperSandro2000 (approved):
LGTM

Do we have a section about those headers in the keppel or cloud docs? If so we should mention this change there to prevent support requests why the header is missing on certain images and that others need to expect that the header could be missing.

majewsky:
I don't see anything in `docs/`, and I've certainly not documented this in the CCloud docs, so we should be good on that front. We can document this in the future, but for now, I'm fine with this flying below the radar.

PR #40, opened 2022-04-11
Add make target for oci-conformance test

SuperSandro2000:
Add make target to generate dummy key, prepare database and start server.
Move storage and auth test driver to trivial to be used for oci conformance.


internal/drivers/trivial/auth.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func getPassword() string {
+	password := os.Getenv("KEPPEL_USERNAME")
+	if password == "" {
+		logg.Fatal("KEPPEL_USERNAME env cannot be empty when using trivial auth")
+	}
+	return password
+}
+
```
SuperSandro2000, line 68:
```suggestion
```

internal/drivers/trivial/auth.go
```diff
@@ -0,0 +1,150 @@
...
+package trivial
+
+import (
+	"encoding/json"
+	"net/http"
+	"os"
+
+	"github.com/go-redis/redis"
+	"github.com/sapcc/go-bits/audittools"
+	"github.com/sapcc/go-bits/logg"
```
SuperSandro2000, line 28:
```suggestion
```

Makefile
```diff
@@ -22,6 +22,16 @@ run-api: build/keppel
 copy-fixtures:
 	find -name '*.actual' | xargs -I{} bash -c 'mv {} $$(echo {} | sed "s/.actual//g")'
 
+privkey.pem:
+	openssl genrsa -out privkey.pem 4096
+
+OCI_DB_CMD := psql -U postgres -d keppel -h 127.0.0.1 -p 54321 -c "INSERT INTO accounts (name, auth_tenant_id) VALUES ('conformance-test', 'bogus') ON CONFLICT DO NOTHING;" -c "INSERT INTO quotas (auth_tenant_id, manifests) VALUES ('bogus', 100) ON CONFLICT DO NOTHING;"
+run-oci-conformance: build/keppel privkey.pem
+# little hack to populare the database
```
SuperSandro2000, line 30:
```suggestion
# little hack to populate the database
```

Makefile.maker.yaml
```diff
@@ -46,3 +46,13 @@ verbatim: |
 
   copy-fixtures:
     find -name '*.actual' | xargs -I{} bash -c 'mv {} $$(echo {} | sed "s/.actual//g")'
+
+  privkey.pem:
+    openssl genrsa -out privkey.pem 4096
+
+  OCI_DB_CMD := psql -U postgres -d keppel -h 127.0.0.1 -p 54321 -c "INSERT INTO accounts (name, auth_tenant_id) VALUES ('conformance-test', 'bogus') ON CONFLICT DO NOTHING;" -c "INSERT INTO quotas (auth_tenant_id, manifests) VALUES ('bogus', 100) ON CONFLICT DO NOTHING;"
+  run-oci-conformance: build/keppel privkey.pem
+  # little hack to populare the database
```
SuperSandro2000, line 55:
```suggestion
  # little hack to populate the database
```

.env-oci-conformance
```diff
@@ -0,0 +1,14 @@
+export KEPPEL_API_PUBLIC_FQDN=localhost
+export KEPPEL_ISSUER_KEY=./privkey.pem
+export KEPPEL_DB_CONNECTION_OPTIONS=sslmode=disable
+export KEPPEL_DB_PASSWORD=mysecretpassword
+export KEPPEL_DB_PORT=54321
+export KEPPEL_USERNAME=keppel_e2e@default/master@ccadmin
+export KEPPEL_PASSWORD=37aded545238cd2414e25f5452322de85f7ca03863c4d09154
```
majewsky, line 7: Since the `trivial` auth driver allows any values for this, we should use more dummy-esque values, for example:

```suggestion
export KEPPEL_USERNAME=johndoe
export KEPPEL_PASSWORD=5uper5ecret
```

Makefile
```diff
@@ -22,12 +21,24 @@ run-api: build/keppel
 copy-fixtures:
 	find -name '*.actual' | xargs -I{} bash -c 'mv {} $$(echo {} | sed "s/.actual//g")'
 
-build-all: build/keppel
+privkey.pem:
+	openssl genrsa -out privkey.pem 4096
```
majewsky, line 25: Since this private key is specifically used for `run-oci-conformance`, the name should clarify that. I suggest putting both this file and also `.env-oci-conformance` into a separate subdirectory, e.g.

```
testing/oci-conformance/privkey.pem
testing/oci-conformance/env.sh
```

Makefile
```diff
@@ -22,12 +21,24 @@ run-api: build/keppel
...
+
+OCI_DB_CMD := psql -U postgres -d keppel -h 127.0.0.1 -p 54321 -c "INSERT INTO accounts (name, auth_tenant_id) VALUES ('conformance-test', 'bogus') ON CONFLICT DO NOTHING;" -c "INSERT INTO quotas (auth_tenant_id, manifests) VALUES ('bogus', 100) ON CONFLICT DO NOTHING;"
+run-oci-conformance: build/keppel privkey.pem
+# little hack to populate the database
+	$(OCI_DB_CMD) || (set -euo pipefail && source ./.env-oci-conformance && timeout 4 $(CURDIR)/build/keppel server api || true) && $(OCI_DB_CMD)
+	@echo "Ready to run distribution-spec"
+	set -euo pipefail && source ./.env-oci-conformance && $(CURDIR)/build/keppel server api
+
+export KEPPEL_USERNAME=dummy-username
```
majewsky, line 34: Why is this needed in the Makefile when `.env-oci-conformance` overrides it?

internal/drivers/trivial/auth.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (d *AuthDriver) AuthenticateUser(userName, password string) (keppel.UserIdentity, *keppel.RegistryV2Error) {
+	expectedUserName := os.Getenv("KEPPEL_USERNAME")
+	if expectedUserName == "" {
+		return nil, keppel.ErrUnauthorized.With(`KEPPEL_USERNAME env cannot be empty when using trivial auth`)
+	}
+	expectedPassword := os.Getenv("KEPPEL_PASSWORD")
+	if expectedPassword == "" {
+		return nil, keppel.ErrUnauthorized.With(`KEPPEL_PASSWORD env cannot be empty when using trivial auth`)
+	}
```
majewsky, line 121: Since environment variables do not usually change after a program has started, I consider it better style to read them only once during initialization, as do all the other drivers that take configuration from environment variables. This would also avoid the code duplication that currently exists between the different methods of this AuthDriver.

Makefile
majewsky: Oh, I'm guessing this is related to the usage of `DummyUserInfo` within the test auth driver? If so, that's spooky action at a distance and not good style. It would be much clearer if the `DummyUserInfo` carries the username in its own struct instead.

majewsky (changes requested):

majewsky (approved):
Rather than go through another round of reviews, I've put some changes on top. Please have a look at the commits that I added.

PR #62, opened 2022-04-27
Allow CMD and ENTYPOINT in Dockerfiles to be a string

SuperSandro2000:


majewsky (changes requested):
We need a solution with less dependencies.

majewsky:
I'd rather we go back to how it used to be before this change:

https://github.com/sapcc/keppel/commit/6dd761b074f#diff-871998b90e78fc50599ad7788a53f52ca6fa6b7faffa08cf3a65449ee45794b2L344-R348

And just use a bespoke type to get the few fields that we actually care about. We don't even want to read the command field at all.

PR #63, opened 2022-05-05
Update redis to v8

SuperSandro2000:
Biggest change for us is that the rate limiter now longer gives out the first access (request/byte) for free hence the changes in the expected test values.

closes #51


internal/drivers/basic/ratelimit.go
```diff
@@ -50,16 +50,16 @@ var (
 		keppel.AnycastBlobBytePullAction: {"KEPPEL_RATELIMIT_ANYCAST_BLOB_PULL_BYTES", "KEPPEL_BURST_ANYCAST_BLOB_PULL_BYTES"},
 	}
 	valueRx          = regexp.MustCompile(`^\s*([0-9]+)\s*[Br]/([smh])\s*$`)
-	rateConstructors = map[string]func(int) throttled.Rate{
-		"s": throttled.PerSec,
-		"m": throttled.PerMin,
-		"h": throttled.PerHour,
+	rateConstructors = map[string]func(int) redis_rate.Limit{
```
majewsky, line 53: Since this now constructs `Limit` instances, it should be called `limitConstructors`.

internal/keppel/ratelimit_driver.go
```diff
@@ -82,28 +84,28 @@ func RegisterRateLimitDriver(name string, factory func(AuthDriver, Configuration
 //implementation.
 type RateLimitEngine struct {
 	Driver RateLimitDriver
-	Store  throttled.GCRAStore
+	Store  *redis.Client
```
majewsky, line 87: By the same argument, this should be called `Client` instead.

internal/keppel/ratelimit_driver.go
```diff
@@ -82,28 +84,28 @@ func RegisterRateLimitDriver(name string, factory func(AuthDriver, Configuration
...
 //the account's rate limit.
-func (e RateLimitEngine) RateLimitAllows(account Account, action RateLimitedAction, amount uint64) (bool, throttled.RateLimitResult, error) {
+func (e RateLimitEngine) RateLimitAllows(account Account, action RateLimitedAction, amount uint64) (bool, *redis_rate.Result, error) {
 	rateQuota := e.Driver.GetRateLimit(account, action)
 	if rateQuota == nil {
 		//no rate limit for this account and action
-		return true, throttled.RateLimitResult{
-			Limit:      math.MaxInt64,
+		return true, &redis_rate.Result{
+			Limit:      redis_rate.Limit{Rate: math.MaxInt64},
```
majewsky, line 97: `Limit` does not make sense without an interval, so we should set it to be safe.
```suggestion
			Limit:      redis_rate.Limit{Rate: math.MaxInt64, Period: time.Second},
```

majewsky (approved):
As a general note, when I approve while requesting changes in comments, this means you have my approval as soon as the changes have been addressed, without requiring an additional round of review.

The logic looks sound to me now. I would like you to adjust some variable names to better line up with what those variables now contain:

PR #74, opened 2022-05-16
Layer decompress to big

SuperSandro2000:


internal/tasks/manifests.go
```diff
@@ -598,6 +605,37 @@ func (j *Janitor) doVulnerabilityCheck(account keppel.Account, repo keppel.Repos
...
+			numberBytes, err := io.Copy(io.Discard, gzipReader)
+			if err != nil {
+				return err
+			}
+
+			b2p := func(b bool) *bool {
+				return &b
+			}
+			// mark blocked for vulnerability scanning if one layer/blob is bigger than 10 GiB
+			blob.BlocksVulnScanning = b2p(numberBytes >= int64(1<<30*blobUncompressedSizeToBigGib))
```
majewsky, line 630: Since `b2p` is only once, we can do without an anon function.
```suggestion
			// mark blocked for vulnerability scanning if one layer/blob is bigger than 10 GiB
			blocksVulnScanning := b2p(numberBytes >= int64(1<<30*blobUncompressedSizeToBigGib))
			blob.BlocksVulnScanning = &blocksVulnScanning
```

internal/tasks/manifests.go
```diff
@@ -598,6 +605,37 @@ func (j *Janitor) doVulnerabilityCheck(account keppel.Account, repo keppel.Repos
...
+		}
+
+		if blob.BlocksVulnScanning != nil && *blob.BlocksVulnScanning {
+			manifest.VulnerabilityStatus = clair.UnsupportedVulnerabilityStatus
+			manifest.VulnerabilityScanErrorMessage = fmt.Sprintf("vulnerability scanning is not supported for uncompressed blobs above %g GiB", blobUncompressedSizeToBigGib)
```
majewsky, line 635:
```suggestion
			manifest.VulnerabilityScanErrorMessage = fmt.Sprintf("vulnerability scanning is not supported for uncompressed image layers above %g GiB", blobUncompressedSizeToBigGib)
```

internal/tasks/manifests.go
```diff
@@ -556,6 +558,11 @@ func (j *Janitor) CheckVulnerabilitiesForNextManifest() (returnErr error) {
 	return err
 }
 
+var (
+	manifestSizeToBigGiB         float64 = 5
+	blobUncompressedSizeToBigGib float64 = 10
```
majewsky, line 563: Typos:
```suggestion
	manifestSizeTooBigGiB         float64 = 5
	blobUncompressedSizeTooBigGib float64 = 10
```

majewsky (commented):
We already discussed this in-person, so I'm just going to leave my suggested edits.

Finally, in `TestCheckVulnerabilitiesForNextManifest`, please add some more comments explaining how the different images are used for different kinds of tests, to avoid future confusion.

majewsky (approved):

PR #116, opened 2022-08-18
Run go-makefile-maker with new lint checks and fix all reported

SuperSandro2000:


internal/test/mock_clair.go
```diff
@@ -98,7 +98,7 @@ func (c *ClairDouble) postIndexReport(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	reqBodyBuf.WriteByte('\n')
-	err = os.WriteFile(actualPathAbs, reqBodyBuf.Bytes(), 0o666)
+	err = os.WriteFile(actualPathAbs, reqBodyBuf.Bytes(), 0o600)
```
majewsky, line 101: This is bad advice. We should use `0o666` and let the user's umask decide which bits to remove.

majewsky (changes requested):

vendor/github.com/sapcc/go-bits/assert/http.go
```diff
@@ -86,6 +86,7 @@ func (r HTTPRequest) Check(t *testing.T, handler http.Handler) (resp *http.Respo
 	handler.ServeHTTP(recorder, request)
 
 	response := recorder.Result()
+	defer response.Body.Close()
```
majewsky, line 89: This needs to be fixed in go-bits.

PR #129, opened 2022-08-30
add the filesystem storage driver

### LINKED ISSUE #122: Local storage driver
Is there a storage driver that could use the local filesystem? Or a remote S3 like one (e.g. minio)?

rgl:
Closes #122

This can be tried/tested with docker compose by checking out the `wip` branch and following https://github.com/rgl/keppel/tree/wip#usage-docker-compose.

### DIFF
--- a/internal/drivers/filesystem/storage.go
+++ b/internal/drivers/filesystem/storage.go
@@ -1,0 +1,267 @@
+/******************************************************************************
+*
+*  Copyright 2022 ruilopes.com
+*
+*  Licensed under the Apache License, Version 2.0 (the "License");
+*  you may not use this file except in compliance with the License.
+*  You may obtain a copy of the License at
+*
+*      http://www.apache.org/licenses/LICENSE-2.0
+*
+*  Unless required by applicable law or agreed to in writing, software
+*  distributed under the License is distributed on an "AS IS" BASIS,
+*  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+*
+******************************************************************************/
+
+package filesystem
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/sapcc/keppel/internal/keppel"
+)
+
+func init() {
+	keppel.RegisterStorageDriver("filesystem", func(_ keppel.AuthDriver, _ keppel.Configuration) (keppel.StorageDriver, error) {
+		rootPath := os.Getenv("KEPPEL_FILESYSTEM_PATH")
+		if rootPath == "" {
+			return nil, fmt.Errorf("the KEPPEL_FILESYSTEM_PATH environment variable must not be empty")
+		}
+		rootPath, err := filepath.Abs(rootPath)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


internal/drivers/filesystem/storage.go
```diff
@@ -0,0 +1,267 @@
...
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/sapcc/keppel/internal/keppel"
+	"github.com/spf13/afero"
```
majewsky, line 33: I generally want to avoid additional library dependencies if the stdlib can serve us instead. I understand the value of the abstractions that afero provides, but in this case we're only really using it to prepend `rootPath` to all paths, and that's something that we could just be doing ourselves in `func blobBasePath` etc. instead.

internal/drivers/filesystem/storage.go
```diff
@@ -0,0 +1,267 @@
...
+*  Unless required by applicable law or agreed to in writing, software
+*  distributed under the License is distributed on an "AS IS" BASIS,
+*  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+*
+******************************************************************************/
+
+// TODO the error returned here are displayed to the user as-is. is that a problem? like, leaking path names?
+// TODO wrap all the errors?
```
majewsky, line 20: I'm mostly fine with leaving errors as-is, but I agree that path names should have the `rootPath` obscured. If you take my other suggestion of using the stdlib functions for filesystem access, that should be easy to add since all those functions always return `*os.PathError` which has the Path in a separate field for easy access and manipulation.

majewsky (changes requested):
Sorry for the delay on my review, I was on vacation last week. I will do one initial review round with just high-level questions, and once these are addressed, I will have a look in more detail.

Besides the notes below, I would like to ask for documentation to be added in `docs/drivers/`, similar to the existing storage drivers.

internal/drivers/filesystem/storage.go
rgl: Using it also makes it possible to run unit tests from a memory filesystem and maybe replacing the trivial driver with this driver.

If that's not relevant at the moment, let me known, and I can replace it with stdlib.

majewsky: Yeah, we aren't going to need that. There is already an in-memory filesystem in the kernel with tmpfs, we don't need to carry code around to simulate it.

rgl: Its now using the stdlib `os` package.

internal/drivers/filesystem/storage.go
rgl: I ended up leaving this as-is.

rgl:
> Besides the notes below, I would like to ask for documentation to be added in `docs/drivers/`, similar to the existing storage drivers.

I've added the documentation. Let me known if it needs more details.

majewsky:
Thank you for your contribution! Instead of going through another round of reviews, I did some minor stylistic adjustments myself in 945705a. I also set up the OCI conformance test to use this storage driver, so we have some automated test coverage going forward. During that, I discovered a bug with multi-chunk uploads that is fixed in 7d78373.

PR #133, opened 2022-09-21
Ignore transinent clair errors and retry scanning

SuperSandro2000:


internal/tasks/manifests_test.go
```diff
@@ -643,4 +643,21 @@ func TestCheckVulnerabilitiesForNextManifest(t *testing.T) {
...
+	simulatedErrorMsg := "failed to scan all layer contents: failed to connect to host=clair-postgresql user=postgres database=clair: dial error (dial tcp 10.30.50.60:5432: connect: connection refused)"
+	mustExec(t, s.DB, `UPDATE manifests SET vuln_status = 'Error', vuln_scan_error = $1 where digest = $2`, simulatedErrorMsg, imageError.Manifest.Digest)
+
+	expectSuccess(t, j.CheckVulnerabilitiesForNextManifest())
+	expectError(t, sql.ErrNoRows.Error(), j.CheckVulnerabilitiesForNextManifest())
+
+	tr.DBChanges().AssertEqualf(`
+		UPDATE blobs SET blocks_vuln_scanning = FALSE WHERE id = 9 AND account_name = 'test1' AND digest = '%[2]s';
+		UPDATE manifests SET next_vuln_check_at = 6120 WHERE repo_id = 1 AND digest = '%[1]s';
+	`, imageError.Manifest.Digest, imageError.Layers[0].Digest)
```
majewsky, line 662: This should verify that the broken index report was deleted in Clair. In fact, something seems to be wrong here. The ClairDouble does not handle DELETE requests, so it should respond 405 (Method Not Allowed) when given a DELETE request. This all smells like the DELETE is not actually taking place.

I'd say you need to:

- extend the mock to handle DELETE requests according to its internal model, and
- have the mock generate the broken index report as a first step, ingest that through the vuln check, than swap out the fixture to generate a valid index report in the second step

majewsky (changes requested):

internal/tasks/fixtures/clair/manifest-004-failure.json
```diff
@@ -0,0 +1,9 @@
+{
```
majewsky, line 1: This file has the same contents as `internal/tasks/fixtures/clair/manifest-004.json` and can be deduplicated.

internal/test/mock_clair.go
```diff
@@ -128,26 +135,53 @@ func (c *ClairDouble) postIndexReport(w http.ResponseWriter, r *http.Request) {
...
+	if c.ReportFixtures[digest] != "" {
+		state = "IndexFinished"
+	}
+	respondwith.JSON(w, http.StatusCreated, map[string]interface{}{
+		"manifest_hash": digest,
+		"state":         state,
+	})
+}
+
+func (c *ClairDouble) deleteIndexReport(w http.ResponseWriter, r *http.Request) {
```
majewsky, line 174: This function should include consistency checks similar to the other ClairDouble request handlers:

- Only expected manifests should be permitted, i.e. there should be an `c.IndexFixtures[digest]`.
- Only manifests that were submitted for indexing should be permitted, i.e. `c.WasIndexSubmitted[digest]` should be true.

majewsky (changes requested):

PR #138, opened 2022-10-06
Fix/dev

Zhurik:
Fixed envs for dev launch + added docker compose

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,6 +20,12 @@ export KEPPEL_DB_CONNECTION_OPTIONS=sslmode=disable
 export KEPPEL_DB_PASSWORD=mysecretpassword
 export KEPPEL_DRIVER_FEDERATION=trivial
 export KEPPEL_DRIVER_INBOUND_CACHE=trivial
+export KEPPEL_API_PUBLIC_FQDN=localhost
+export KEPPEL_DRIVER_AUTH=trivial
+export KEPPEL_USERNAME=johndoe
+export KEPPEL_PASSWORD=SuperSecret
+export KEPPEL_DRIVER_STORAGE=filesystem
+export KEPPEL_FILESYSTEM_PATH=./keppel
 ```
 
 In addition to that the following extra steps are required:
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -20,13 +20,33 @@ export KEPPEL_DB_CONNECTION_OPTIONS=sslmode=disable
...
 - A local postgresql instance. You can start one in docker with the following command: `docker run --name postgres -e POSTGRES_PASSWORD=mysecretpassword -e POSTGRES_DB=keppel -p 127.0.0.1:5432:5432 -d postgres`
 - `KEPPEL_DRIVER_AUTH` and `KEPPEL_DRIVER_STORAGE` need to be configured and exported. Best look them up in an existing Keppel installation.
 
+### Run Keppel with docker-compose
+
+1. Generate private key locally:
+
+```bash
+openssl genrsa -out privkey.pem 4096
+```
```
SuperSandro2000, line 42: This is already in line 32

CONTRIBUTING.md
```diff
@@ -20,13 +20,33 @@ export KEPPEL_DB_CONNECTION_OPTIONS=sslmode=disable
...
+
+2. Run docker-compose
+
+```bash
+docker-compose up
+```
```
SuperSandro2000, line 48: We are using ``bash ./testing/with-postgres-db.sh make run-api`` for local testing, so that is always guaranteed to work. The docker-compose part will likely bit rot a bit over time since it will not be tested by us. Is there a particular problem with the above command?

Zhurik: It's just more convenient to test first things with two commands. If you think that those changes are unnecessary I'll close PR

CONTRIBUTING.md
Zhurik: Just to be sure, I can remove it

CONTRIBUTING.md
SuperSandro2000: I've pushed a commit to update the guide to be fully working again with a sentence on how to use ``./testing/with-postgres-db.sh``. From my perspective it should now be as convenient as using docker-compose plus it is tested by us.

PR #145, opened 2022-11-02
Only submit image layers to clair that it will understand

SuperSandro2000:


majewsky (changes requested):
I know that I advocated for this approach when we talked about this yesterday, but upon considering the implications once again, I have to reverse my stance. With this implementation, any layers not fitting the bill will be silently ignored. So when newer clients start building images with different layer media types, and these get pushed into Keppel, we will always report vulnerability status "Clean", which is sort of a lie. Your first intuition (marking images with unknown layers as vulnerability status "Unsupported") is preferable since that means customers will come to us when they try to push newer formats that we don't send to Clair yet.

PR #156, opened 2022-11-30
Split vurnability check queues to generate better metrics

SuperSandro2000:


internal/tasks/manifests.go
```diff
@@ -532,12 +541,32 @@ func (j *Janitor) CheckVulnerabilitiesForNextManifest() (returnErr error) {
 		}
 	}()
 
-	//find manifest to sync
+	return j.checkVulnerabilitiesForManifest(vulnCheckSelectQuery)
+}
+
+// CheckVulnerabilitiesForNextPendingManifest is very similar to CheckVulnerabilitiesForNextManifest
+// with the exception that it only checks for already to clair submitted manifests which are still
+// in vuln_status Pending
```
majewsky, line 549: It also checks manifests that were never submitted to Clair (when a manifest is first added, it's vuln_status is initialized as "Pending").

internal/tasks/manifests_test.go
```diff
@@ -595,11 +595,11 @@ func TestCheckVulnerabilitiesForNextManifest(t *testing.T) {
...
-	expectSuccess(t, j.CheckVulnerabilitiesForNextManifest())
-	expectSuccess(t, j.CheckVulnerabilitiesForNextManifest())
-	expectSuccess(t, j.CheckVulnerabilitiesForNextManifest())
-	expectSuccess(t, j.CheckVulnerabilitiesForNextManifest())
+	expectSuccess(t, j.CheckVulnerabilitiesForNextPendingManifest()) //once for each manifest
+	expectSuccess(t, j.CheckVulnerabilitiesForNextPendingManifest())
+	expectSuccess(t, j.CheckVulnerabilitiesForNextPendingManifest())
+	expectSuccess(t, j.CheckVulnerabilitiesForNextPendingManifest())
+	expectSuccess(t, j.CheckVulnerabilitiesForNextPendingManifest())
 	expectError(t, sql.ErrNoRows.Error(), j.CheckVulnerabilitiesForNextManifest())
```
majewsky, line 603: The tests should cover that both job loops eventually enter `sql.ErrNoRows`. For example, in this case, we expect:

- that `CheckVulnerabilitiesForNextManifest` returns `sql.ErrNoRows` right away
- that `CheckVulnerabilitiesForNextPendingManifest` succeeds five times and then returns `sql.ErrNoRows` after that

Same for the other instances of the same pattern below. (For some of them, you already changed it in the right way.)

internal/tasks/metrics.go
```diff
@@ -41,6 +41,14 @@ var (
...
 		Help: "Counter for vulnerability checks that were retried due to transient errors in Clair.",
 	})
+	checkPendingVulnerabilitySuccessCounter = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "keppel_successful_pending_vulnerability_checks",
+		Help: "Counter for successful updates of the pending vulnerability status of a manifest.",
+	})
+	checkPendingVulnerabilityFailedCounter = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "keppel_failed_pending_vulnerability_checks",
+		Help: "Counter for failed updates of the pending vulnerability status of a manifest.",
+	})
```
majewsky, line 51: I don't like these metric names. It's not a "pending vulnerability check", it's a vulnerability check for an image in vulnerability status "Pending". I don't have a good idea for a succinct metric name right now; just want to register my dissatisfaction with these names for the moment.

majewsky (changes requested):

internal/tasks/manifests_test.go
SuperSandro2000: I hope I catched all of them

majewsky:
As discussed on the phone today, we will allocate time for a more comprehensive rewrite of the Clair integration that will enable arbitrary horizontal scaling of the vulnerability check jobloop.

PR #160, opened 2022-12-14
Split vulnerability columns into their own table

SuperSandro2000:


majewsky:
@SuperSandro2000 asked me to take a look why `TestCheckVulnerabilitiesForNextManifestWithError` is failing right now.

The root cause is that the test does not step the clock after `image.MustUpload`. Since the `vulnCheckSelectQuery` looks for vuln_info entries with `v.next_check_at < $1` (where `$1` is the current time), the uploaded image does not match. I suggest changing this to `v.next_check_at <= $1`.

majewsky (changes requested):
My overall impression is good, tough I do have a fair amount of notes. As usual, this is the broad sweep. I will do another more detailed review once these points are addressed.

internal/keppel/models.go
```diff
@@ -461,6 +458,43 @@ type UnknownManifest struct {
...
+
+	if err == sql.ErrNoRows {
+		_, err := db.Exec(`
+				INSERT INTO vuln_info SET repo_id = $1, digest = $2, status = '', message = 'Pending', next_check_at = $3`,
+			id, manifestDigest, time.Now(),
+		)
+		if err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
SuperSandro2000: Turns out I already fixed this and the code is no longer needed.

SuperSandro2000:
Fixed all the comments you had.

internal/tasks/manifests_test.go
```diff
@@ -690,8 +693,8 @@ func TestCheckVulnerabilitiesForNextManifestWithError(t *testing.T) {
 	expectSuccess(t, j.CheckVulnerabilitiesForNextManifest())
 	expectError(t, sql.ErrNoRows.Error(), j.CheckVulnerabilitiesForNextManifest())
 	tr.DBChanges().AssertEqualf(`
-		UPDATE manifests SET next_vuln_check_at = %[2]d WHERE repo_id = 1 AND digest = '%[1]s';
-	`, image.Manifest.Digest, s.Clock.Now().Add(2*time.Minute).Unix())
+		UPDATE vuln_info SET next_check_at = %[3]d WHERE repo_id = 1 AND digest = '%[2]s';
+	`, image.Layers[0].Digest, image.Manifest.Digest, s.Clock.Now().Add(2*time.Minute).Unix())
```
majewsky, line 697: `%[1]s` is never used here, so it can be removed from the arg list.

internal/api/keppel/fixtures/delete-account-000.sql
```diff
@@ -18,8 +18,8 @@ INSERT INTO manifest_contents (repo_id, digest, content) VALUES (1, 'sha256:6aa9
...
+INSERT INTO manifests (repo_id, digest, media_type, size_bytes, pushed_at, validated_at, validation_error_message, last_pulled_at, labels_json, gc_status_json, min_layer_created_at, max_layer_created_at) VALUES (1, 'sha256:6aa9f3d5659c999fecab6df26efb864792763a2c7ae7580edf5dc11df2882ea5', 'application/vnd.docker.distribution.manifest.list.v2+json', 909, 0, 0, '', NULL, '', '', NULL, NULL);
+INSERT INTO manifests (repo_id, digest, media_type, size_bytes, pushed_at, validated_at, validation_error_message, last_pulled_at, labels_json, gc_status_json, min_layer_created_at, max_layer_created_at) VALUES (1, 'sha256:e255ca60e7cfef94adfcd95d78f1eb44404c4f5887cbf506dd5799489a42606c', 'application/vnd.docker.distribution.manifest.v2+json', 592, 100, 100, '', NULL, '', '', NULL, NULL);
```
majewsky, line 22: More manifests without vuln_info.

majewsky (changes requested):
The fine sweep.

majewsky (approved):

PR #175, opened 2023-03-02
Make Swift containers write restricted

talal:


internal/drivers/openstack/federation_swift.go
```diff
@@ -90,7 +90,12 @@ func initSwiftContainerConnection(envPrefix string) (*schwift.Container, error)
 	if err != nil {
 		return nil, err
 	}
-	container, err := swiftAccount.Container(osext.MustGetenv(envPrefix + "SWIFT_CONTAINER")).EnsureExists()
+	container := swiftAccount.Container(osext.MustGetenv(envPrefix + "SWIFT_CONTAINER"))
+	//add the 'X-Container-Meta-Write-Restricted' metadata header to restrict writes only
+	//to allowed users. See: https://github.com/sapcc/swift-addons/tree/master#write-restriction
+	hdr := schwift.NewContainerHeaders()
+	hdr.Metadata().Set("Write-Restricted", "true")
+	err = container.Create(hdr.ToOpts())
```
majewsky, line 98: We don't need write restriction on the federation DB. This container is not located in a customer account.

majewsky (changes requested):

majewsky (approved):

PR #180, opened 2023-03-10
clair: fill index_started_at and index_finished_at

SuperSandro2000:


internal/tasks/manifests.go
```diff
@@ -711,12 +711,14 @@ func (j *Janitor) doVulnerabilityCheck(account keppel.Account, repo keppel.Repos
...
 	if len(layerBlobs) > 0 {
-		clairState, err := j.cfg.ClairClient.CheckManifestState(manifest.Digest, func() (clair.Manifest, error) {
+		clairState, err := j.cfg.ClairClient.CheckManifestState(manifest.Digest, j.timeNow, func() (clair.Manifest, error) {
 			return j.buildClairManifest(account, manifest, layerBlobs)
 		})
 		if err != nil {
 			return err
 		}
+		vulnInfo.IndexStartedAt = clairState.IndexingStartedAt
+		vulnInfo.IndexFinishedAt = clairState.IndexingFinishedAt
```
majewsky, line 721: After now reviewing the relevant code in its entirety, I don't think we need to change `CheckManifestState` at all. Please correct me if I'm wrong, but as far as I can see:

- `vulnInfo.IndexStartedAt` needs to be filled here if it is currently nil **or** `clairState.IndexingWasRestarted`.
- `vulnInfo.IndexFinishedAt` needs to be filled here if it is currently nil **and** `clairState.IsIndexed`.

SuperSandro2000: 🤔

internal/tasks/manifests.go
```diff
@@ -733,6 +738,8 @@ func (j *Janitor) doVulnerabilityCheck(account keppel.Account, repo keppel.Repos
 				return fmt.Errorf("Clair reports indexing of %s as finished, but vulnerability report is 404", manifest.Digest)
 			}
 			vulnStatuses = append(vulnStatuses, clairReport.VulnerabilityStatus())
+			now := j.timeNow()
+			vulnInfo.IndexFinishedAt = &now
```
majewsky, line 742: This should only be set if the field is nil before, otherwise it will just report when we last checked the vulnerability status.

I was expecting this to show up in a test, but if I'm not mistaken, we don't have a test where an image is already indexed and then the vulnerability status changes in Clair. Please add such a test, or extend an existing test.

majewsky (changes requested):

internal/clair/index.go
```diff
@@ -165,9 +174,30 @@ func (c *Client) submitManifest(renderManifest func() (Manifest, error)) (indexR
...
 	}
-	var result indexReport
-	err = c.doRequest(req, &result)
-	return result, err
+
+	// get and return index state hash to later resubmit reports if the configuration changed
+	req, err = http.NewRequest(
+		http.MethodGet,
+		c.requestURL("indexer", "api", "v1", "index_state"),
+		bytes.NewReader(jsonBytes),
```
majewsky, line 190: GET requests don't accept bodies.
```suggestion
		nil,
```

internal/tasks/manifests.go
```diff
@@ -733,6 +740,11 @@ func (j *Janitor) doVulnerabilityCheck(account keppel.Account, repo keppel.Repos
 				return fmt.Errorf("Clair reports indexing of %s as finished, but vulnerability report is 404", manifest.Digest)
 			}
 			vulnStatuses = append(vulnStatuses, clairReport.VulnerabilityStatus())
+
+			if vulnInfo.IndexFinishedAt == nil {
+				now := j.timeNow()
```
majewsky, line 745: We already knew `IsIndexed` before starting `GetVulnerabilityReport`, so we can use the same `now` as before.

PR #181, opened 2023-03-10
refactor Accept handler handling to reduce cognitive complexity

majewsky:
The special case rules get moved into the ParsedManifest interface. According to gocognit, this reduces the complexity of handleGetOrHeadManifest() from 100 to 66. I also used the opportunity to turn an incorrect 404 response into the more appropriate 406 (Not Acceptable).

A change was needed in the RepoClient since my understanding of the Accept header was actually incorrect. It's not allowed to have multiple Accept headers. Instead, all acceptable MIME types, must be listed in a single header.

### DIFF
--- a/internal/keppel/manifest.go
+++ b/internal/keppel/manifest.go
@@ -59,6 +59,11 @@ type ParsedManifest interface {
 	BlobReferences() []distribution.Descriptor
 	//ManifestReferences returns all manifests referenced by this manifest.
 	ManifestReferences(pf PlatformFilter) []manifestlist.ManifestDescriptor
+	//AcceptableAlternates returns the subset of ManifestReferences() that is
+	//acceptable as alternate representations of this manifest. When a client
+	//asks for this manifest, but the Accept header does not match the manifest
+	//itself, the API will look for an acceptable alternate to serve instead.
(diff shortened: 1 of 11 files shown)


internal/keppel/manifest.go
```diff
@@ -147,3 +160,19 @@ func (a listManifestAdapter) ManifestReferences(pf PlatformFilter) []manifestlis
 	}
 	return result
 }
+
+func (a listManifestAdapter) AcceptableAlternates(pf PlatformFilter) []manifestlist.ManifestDescriptor {
+	var result []manifestlist.ManifestDescriptor
+	for _, m := range a.ManifestReferences(pf) {
+		//If we have an application/vnd.docker.distribution.manifest.list.v2+json manifest, but the
+		//client only accepts application/vnd.docker.distribution.manifest.v2+json, in order to stay
```
SuperSandro2000, line 168: Do we know any client that does that? Why not check if the list contains only one entry and choose that?

majewsky: This is relevant for `docker pull`. The "only one entry" case does not apply because this is specifically about multi-arch images.

SuperSandro2000 (approved):

PR #183, opened 2023-03-14
add internal/client/peer package

majewsky:
We have various requests to the peer API that all play out roughly the same, but have their HTTP request processing code duplicated at each usage site. This also breaks my favorite design principle of "stay on one level of thought". This PR centralizes the handling of these requests into `package peerclient` with a more high-level API.

This reduces the gocognit score of `func handlePutAccount` from 163 to 142.


internal/processor/manifests.go
```diff
@@ -809,43 +810,17 @@ func (p *Processor) downloadManifestViaPullDelegation(imageRef keppel.ImageRefer
 		return nil, "", false
 	}
 
-	//get token for peer
-	peerToken, err := auth.GetPeerToken(p.cfg, peer, auth.PeerAPIScope)
+	peerClient, err := peerclient.New(p.cfg, peer, auth.PeerAPIScope)
 	if err != nil {
-		logg.Error("while trying to get a peer token for pull delegation: %s", err.Error())
+		logg.Error(err.Error())
```
SuperSandro2000, line 815: Don't we loose the error location here?

internal/client/peer/client.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	err = json.Unmarshal(respBodyBytes, &data)
+	if err != nil {
+		return "", fmt.Errorf("could not get peer token: %w", err)
+	}
+	return data.Token, nil
+}
+
+func (c Client) buildRequestURL(path string) string {
+	return fmt.Sprintf("https://%s/%s", c.peer.HostName, path)
```
SuperSandro2000, line 82: We could use this in line 47, too

majewsky: Done in f96f151.

internal/processor/manifests.go
majewsky: We do. Fixed during f96f151.

PR #184, opened 2023-03-15
Fill in checked_at and check_duration_secs

SuperSandro2000:


internal/tasks/manifests.go
```diff
@@ -615,6 +621,8 @@ func (j *Janitor) doVulnerabilityCheck(account keppel.Account, repo keppel.Repos
 		return nil
 	}
 
+	vulnInfo.CheckStartedAt = j.timeNow()
```
majewsky, line 624: This can also be set right before calling `doVulnerabilityCheck()`, and then it can be a method-local variable instead of appearing in the DB model type.

majewsky (changes requested):

majewsky:
We refactored the PR in a pair programming session.

PR #191, opened 2023-03-21
Add new job loop to check index_state and if it changed slowly reschule vulnerability reports

SuperSandro2000:


internal/clair/index.go
```diff
@@ -142,6 +141,38 @@ func (c *Client) CheckManifestState(digest string, renderManifest func() (Manife
...
+	req, err := http.NewRequest(http.MethodDelete, c.getIndexReportURL(digest), http.NoBody)
+	if err != nil {
+		return err
+	}
+	err = c.doRequest(req, nil)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
majewsky, line 154: Can be shortened:
```suggestion
	return c.doRequest(req, nil)
```

internal/tasks/manifests.go
```diff
@@ -794,3 +794,82 @@ func (j *Janitor) buildClairManifest(account keppel.Account, manifest keppel.Man
...
+	}
+
+	total, err := j.db.SelectInt("SELECT COUNT (*) FROM vuln_info")
+	if err != nil {
+		return err
+	}
+	inPending, err := j.db.SelectInt(fmt.Sprintf("SELECT COUNT(*) FROM vuln_info WHERE status = '%s'", clair.PendingVulnerabilityStatus))
+	if err != nil {
+		return err
+	}
```
majewsky, line 819: Sequence scans are expensive, so we should do both at once. Something along those lines:

```go
query := `SELECT COUNT(*), COUNT(CASE WHEN status = 'Pending' THEN TRUE ELSE NULL END) FROM vuln_info`
err := j.db.QueryRow(query).Scan(&total, &inPending)
```

internal/tasks/manifests.go
```diff
@@ -794,3 +794,82 @@ func (j *Janitor) buildClairManifest(account keppel.Account, manifest keppel.Man
...
+			}
+
+			err = j.setManifestAndParentsToPending(digest)
+			return err
+		},
+	)
+	return err
+}
+
+func (j *Janitor) setManifestAndParentsToPending(manifestDigest string) error {
```
majewsky, line 849: This entire function can be fused into a single SQL query:
```sql
UPDATE vuln_info SET status = $1, next_check_at = $2
  WHERE digest = $3 OR digest IN (
    SELECT parent_digest FROM manifest_manifest_refs WHERE child_digest = $3
  )
```

internal/tasks/manifests_test.go
```diff
@@ -724,7 +724,21 @@ func TestCheckVulnerabilitiesForNextManifestWithError(t *testing.T) {
...
+	expectSuccess(t, j.CheckClairManifestState())
+	tr.DBChanges().AssertEqualf(`
+		UPDATE vuln_info SET status = '%[3]s', next_check_at = %[2]d WHERE repo_id = 1 AND digest = '%[1]s';
+	`, image.Manifest.Digest, s.Clock.Now().Unix(), clair.PendingVulnerabilityStatus)
+
+	s.Clock.StepBy(1 * time.Minute)
+	expectSuccess(t, ExecuteOne(j.CheckVulnerabilitiesForNextManifest()))
+	expectError(t, sql.ErrNoRows.Error(), ExecuteOne(j.CheckVulnerabilitiesForNextManifest()))
+	tr.DBChanges().AssertEqualf(`
+		UPDATE vuln_info SET status = '%[4]s', next_check_at = %[2]d, checked_at = %[3]d WHERE repo_id = 1 AND digest = '%[1]s';
```
majewsky, line 742: The `index_state` column for this `vuln_info` is filled in line 704, but then changed in the Clair mock in line 732. Should it not be updated to match the new value here?

SuperSandro2000: Good catch! One query was missing updating a value.

majewsky (approved):

PR #195, opened 2023-03-30
Reduce cognitive code complexity for executeGCPolicies

SuperSandro2000:


internal/tasks/image_gc.go
```diff
@@ -194,91 +200,102 @@ func (j *Janitor) executeGCPolicies(account keppel.Account, repo keppel.Reposito
...
-		//evaluate policy for each manifest
-		for _, m := range manifests {
-			//skip those manifests that are already deleted, and those which are
-			//protected by an earlier policy or one of the baseline checks above
-			if m.IsDeleted || m.GCStatus.IsProtected() {
-				continue
-			}
+	//finalize and persist GCStatus for all affected manifests
+	query = `UPDATE manifests SET gc_status_json = $1 WHERE repo_id = $2 AND digest = $3`
+	err = sqlext.WithPreparedStatement(j.db, query, manifests.persistGCStatus)
```
majewsky, line 214: The query string is tightly coupled to the behavior inside `persistGCStatus` and should be moved into that function. Also, I would suggest to have this be a function on `j`, in which case we would not need a type declaration for `manifestDataSlice`.
```suggestion
	err = j.persistGCStatus(manifests)
```

majewsky (changes requested):

internal/tasks/image_gc.go
SuperSandro2000: The type is still used in evaluatePolicy()

majewsky: I don't think it is? evaluatePolicy takes a `[]*manifestData` instead of a `manifestDataSlice`.

majewsky (approved):

PR #197, opened 2023-03-30
Refactor doRequest to reduce cognitive complexity

SuperSandro2000:


majewsky:
I have a counterproposal in #206. I do not like this change because it takes out blocks that don't make that much sense on their own: `doAuthChallenge` also resends the request after the reauth, contrary to what the name suggests. And `tryToParseAndProxyError` makes a lot of implicit assumptions about the situation in which it will be called, so splitting it out into a separate function makes it harder to read than leaving it inline in my opinion.

PR #205, opened 2023-04-11
Replace crypto/sha256 with minio/sha256-simd

SuperSandro2000:
before:
$ time bash ./testing/with-postgres-db.sh make build/cover.html
real    0m37.439s
user    0m28.226s
sys     0m5.689s

after:
$ time bash ./testing/with-postgres-db.sh make build/cover.html
real    0m27.799s
user    0m15.087s
sys     0m3.690s


majewsky:
I don't manage imports manually most of the time, so I can foresee myself just typing `sha256.New()` or such in a new file later on and goimports decides to pick `crypto/sha256` instead of the SIMD variant without me noticing. Could you please add a lint for this? golangci-lint supports https://github.com/OpenPeeDeeP/depguard which should be able to do this from a quick glance.

PR #219, opened 2023-05-08
Misc cleanup

SuperSandro2000:


majewsky:
It's not letting me post on all lines for some reason, but hopefully this works. In `internal/tasks/manifests.go:423`:

> Typo: `shell` -> `shall`
> 
> In fact, I would suggest the name `digestToBeDeleted` instead. The map is named with `shall` because it is boolean-valued.

majewsky (approved):

PR #220, opened 2023-05-08
Add Trivy jobloop

SuperSandro2000:
depends on https://github.com/sapcc/go-bits/pull/77


internal/tasks/manifests.go
```diff
@@ -710,11 +710,11 @@ func (j *Janitor) doVulnerabilityCheck(account keppel.Account, repo keppel.Repos
 		return nil
 	}
 
-	layerBlobs, continueCheck, err := j.checkPreConditionsForClair(account, repo, manifest, vulnInfo)
+	layerBlobs, err := j.checkPreConditionsForClair(account, repo, manifest, vulnInfo)
 	if err != nil {
 		return err
 	}
-	if !continueCheck {
+	if layerBlobs == nil {
```
majewsky, line 720: What happens if we return true together with an empty list of blobs? In that case, this would be a behavior change. Can this case happen?

SuperSandro2000: 🤔 so we should rather do something like:
```suggestion
	if len(layerBlobs) == 0 {
```

SuperSandro2000: fixed in #219

internal/tasks/manifests.go
```diff
@@ -891,3 +894,160 @@ func (j *Janitor) setManifestAndParentsToPending(ctx context.Context, manifestDi
...
+	securityInfo.Message = "" //unless it gets set to something else below
+	imageRef := keppel.ImageReference{
+		Host:      j.cfg.APIPublicHostname,
+		RepoName:  repo.Name,
+		Reference: keppel.ManifestReference{Digest: manifest.Digest},
+	}
+	requestURL := j.cfg.Trivy.URL
+	requestURL.Path = "/trivy"
+	requestURL.RawQuery = "image=" + imageRef.String()
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, requestURL.String(), http.NoBody)
```
majewsky, line 1000: That's a response, not a request.
```suggestion
	resp, err := http.NewRequestWithContext(ctx, http.MethodGet, requestURL.String(), http.NoBody)
```

SuperSandro2000: The request is only done in the code I add in the next force push.

cmd/trivyproxy/main.go
```diff
@@ -108,7 +109,9 @@ func (a *API) proxyToTrivy(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	stdout, stderr, err := a.runTrivy(r.Context(), imageURL)
+	keppelToken := r.Header.Get(trivy.KeppelToken)
```
majewsky, line 112: This does not read very well: `trivy.KeppelToken` sounds like it would contain a token value, instead of the name of the token header.

```suggestion
	keppelToken := r.Header.Get(trivy.KeppelTokenHeader)
```

cmd/janitor/main.go
```diff
@@ -83,6 +83,9 @@ func run(cmd *cobra.Command, args []string) {
 	if cfg.ClairClient != nil {
 		go tasks.GoQueuedJobLoop(ctx, 3, janitor.CheckVulnerabilitiesForNextManifest())
 	}
+	if cfg.Trivy != nil {
+		go janitor.CheckTrivySecurityStatus(nil).Run(ctx)
```
majewsky, line 87: We should at least make a note here that this needs to be scaled to multiple goroutines eventually.

internal/tasks/manifests.go
```diff
@@ -891,3 +893,178 @@ func (j *Janitor) setManifestAndParentsToPending(ctx context.Context, manifestDi
...
+		},
+		ProcessRow: j.processTrivySecurityInfo,
+	}).Setup(registerer)
+}
+
+func (j *Janitor) processTrivySecurityInfo(tx *gorp.Transaction, securityInfo keppel.TrivySecurityInfo, labels prometheus.Labels) error {
+	//load corresponding repo, account and manifest
+	repo, err := keppel.FindRepositoryByID(tx, securityInfo.RepositoryID)
+	if err != nil {
+		return fmt.Errorf("cannot find repo for manifest %s: %s", securityInfo.Digest, err.Error())
```
majewsky, line 930: When an underlying error is wrapped, %w should be used to preserve the inner error type.

```suggestion
		return fmt.Errorf("cannot find repo for manifest %s: %w", securityInfo.Digest, err)
```

Same for the other error returns in this method.

internal/tasks/manifests.go
```diff
@@ -891,3 +893,178 @@ func (j *Janitor) setManifestAndParentsToPending(ctx context.Context, manifestDi
...
+		return err
+	}
+
+	var securityStatuses []clair.VulnerabilityStatus
+	for _, result := range trivyReport.Results {
+		for _, vulnerability := range result.Vulnerabilities {
+			if securityStatus, ok := clair.MapToTrivySeverity[vulnerability.Severity]; ok {
+				securityStatuses = append(securityStatuses, securityStatus)
+			} else {
+				return fmt.Errorf("vulnerability severity with name %s returned from trivy is unknown and cannot be mapped", securityStatus)
```
majewsky, line 1019: This path does not update `NextCheckAt`, so the jobloop will immediately retry the same image and get stuck on it.

Since you already have a defer that updates some metadata fields on success, I recommend using that function to also update metadata in the error case there.

internal/tasks/manifests.go
```diff
@@ -891,3 +893,178 @@ func (j *Janitor) setManifestAndParentsToPending(ctx context.Context, manifestDi
...
+			ResourceName: repo.FullName(),
+			Actions:      []string{"pull"},
+		}),
+	}.IssueTokenWithExpires(j.cfg, 20*time.Minute)
+	if err != nil {
+		return err
+	}
+
+	//ask Trivy for the security status of the manifest
+	securityInfo.Message = "" //unless it gets set to something else below
```
majewsky, line 1006: The Message field does not get populated on errors. Tying into my previous review comment, I suggest filling Message on error in the deferred function:

```go
if err != nil {
  securityInfo.Status = clair.ErrorVulnerabilityStatus
  securityInfo.Message = err.Error()
  securityInfo.NextCheckAt = j.timeNow().Add(j.addJitter(5 * time.Minute))
}
```

internal/tasks/manifests_test.go
```diff
@@ -547,24 +547,24 @@ func answerMostWith404(h http.Handler) http.HandlerFunc {
...
+		j, s := setup(t, test.WithClairDouble, test.WithTrivyDouble)
 		s.Clock.StepBy(1 * time.Hour)
 
 		//setup two image manifests with just one content layer (we don't really care about
 		//the content since our Clair double doesn't care either)
 		images := make([]test.Image, 3)
 		for idx := range images {
 			images[idx] = test.GenerateImage(test.GenerateExampleLayer(int64(idx)))
-			images[idx].MustUpload(t, s, fooRepoRef, "")
+			images[idx].MustUpload(t, s, fooRepoRef, fmt.Sprint(idx))
```
majewsky, line 558: While not wrong, I found this choice of function surprising. A more conventional choice is `Itoa`.
```suggestion
			images[idx].MustUpload(t, s, fooRepoRef, strconv.Itoa(idx))
```

majewsky, line 558: Also, on a more general note, why do we need tags in this testcase? Trivy should only see digest refs like `keppel.example.com/account/repo@sha256:a1b2c3...` anyway, no?

internal/test/mock_trivy.go
```diff
@@ -0,0 +1,78 @@
...
+	"os"
+	"testing"
+
+	"github.com/gorilla/mux"
+	"github.com/sapcc/go-bits/httpapi"
+	"github.com/sapcc/go-bits/respondwith"
+)
+
+// ClairDouble acts as a test double for a Clair API.
+type TrivyDouble struct {
```
majewsky, line 35: Copy-paste error in the comments. There are more instances of this further below in the same file.

Also, I can't help but nitpick that this is _technically_ not a TrivyDouble, but a TrivyProxyDouble. :)

internal/trivy/trivy.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (tc *Config) ScanManifest(manifestRefString, keppelToken string) (trivyReport securityRespone, returnedError error) {
+	//we don't allow Trivy to take more than 10 minutes on a single image (which is already an
+	//insanely generous timeout)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+	defer cancel()
+
+	requestURL := tc.URL
+	requestURL.Path = "/trivy"
+	requestURL.RawQuery = "image=" + manifestRefString
```
majewsky, line 59: I don't have an immediate reason to suggest that `manifestRefString` could contain special chars that break the query syntax, but it's probably still a good idea to be defensive here.

```suggestion
	requestURL.RawQuery = url.Values{"image": {manifestRefString}}.Encode()
```

majewsky (changes requested):
Lots of comments, but all are rather small ones.

majewsky (approved):

PR #226, opened 2023-05-24
Trivy policies

majewsky:
This is not everything that relates to these policies, but it's a good point to stop enlarging this PR. The remaining part is to include policy-guided assessments in the reports that we forward from Trivy to the user.


docs/api-spec.md
```diff
@@ -351,6 +351,55 @@ token needs to be presented when creating a replica of this primary account.
...
+    {
+      "match_repository": ".*",
+      "only_unfixed": true,
+      "action": {
+        "ignore": true,
+        "assessment": "risk accepted: vulnerabilities without an available fix are not actionable"
+      }
+    },
+    {
+      "match_repository": "my-python-app|my-other-image",
```
SuperSandro2000, line 372: Are those the keppel repos or the trivy "sources" like ``debian:stable (debian 11.7)``?

docs/api-spec.md
```diff
@@ -351,6 +351,55 @@ token needs to be presented when creating a replica of this primary account.
...
+}
+```
+
+The following fields may be returned:
+
+| Field | Type | Explanation |
+| ----- | ---- | ----------- |
+| `policies` | array of objects | Each policy matches a certain set of vulnerabilities on images in a certain set of repositories. Policies can be used to adjust the severity of vulnerabilities or ignore them altogether in certain situations. |
+| `policies[].match_repository` | string | The policy applies to all repositories in this account whose name matches this regex. The leading account name and slash is stripped from the repository name before matching. The notes on regexes below apply. |
+| `policies[].match_vulnerability_id` | string | The policy applies to all vulnerabilities whose ID (as shown in the `VulnerabilityID` field of Trivy's JSON report format) matches this regex. The notes on regexes below apply. |
```
SuperSandro2000, line 389: We should mention that this typically is a CVE or equivalent.

docs/api-spec.md
```diff
@@ -351,6 +351,55 @@ token needs to be presented when creating a replica of this primary account.
...
+| `policies[].only_unfixed` | bool or omitted | If true, the policy applies only to those vulnerabilities for which no fixed version is available (that is, the `FixedVersion` field is missing in Trivy's JSON report format). |
```
SuperSandro2000, line 390:
```suggestion
| `policies[].only_unfixed` | bool or omitted | If true, the policy only applies to those vulnerabilities for which no fixed version is available in the distribution repository (that is, the `FixedVersion` field is missing in Trivy's JSON report format). |
```
We should simplify this sentence. 
We need to be clear, where the fix needs to be. It could be that it is commited, released on pypi but not in the debian package yet or stuck somewhere in the process.

Also maybe we could rename only_unfixed to only_no_update or something similar because there could be a fix available which is not in the form of a package update.

docs/api-spec.md
```diff
@@ -351,6 +351,55 @@ token needs to be presented when creating a replica of this primary account.
...
+| `policies[].action` | object | The effect that this policy will have on matching vulnerabilities reported for images in matching repositories.  |
+| `policies[].action.assessment` | string | A human-readable description of the reasoning behind this policy. |
```
SuperSandro2000, line 392: Should we limit this to some size?

docs/api-spec.md
```diff
@@ -351,6 +351,55 @@ token needs to be presented when creating a replica of this primary account.
...
+| `policies[].action.ignore` | bool or omitted | If true, matching vulnerabilities will be ignored when computing the aggregated vulnerability status of the respective image. This is the same effect as if `action.severity` was set to `Clean`, but the intent is clearer. |
```
SuperSandro2000, line 393: image or manifest?

docs/api-spec.md
```diff
@@ -351,6 +351,55 @@ token needs to be presented when creating a replica of this primary account.
...
+| `policies[].action.severity` | string or omitted | If present, matching vulnerabilities will be treated as having the given severity when computing the aggregated vulnerability status of the respective image. |
```
SuperSandro2000, line 394: We should list the possible values here.

docs/api-spec.md
```diff
@@ -351,6 +351,55 @@ token needs to be presented when creating a replica of this primary account.
...
+      "match_repository": ".*",
+      "only_unfixed": true,
+      "action": {
+        "ignore": true,
+        "assessment": "risk accepted: vulnerabilities without an available fix are not actionable"
+      }
+    },
+    {
+      "match_repository": "my-python-app|my-other-image",
+      "match_vulnerability": "CVE-2022-40897",
```
SuperSandro2000, line 373: Maybe we should add a match severity if it would be a use case to ignore low, unfixed CVEs.

SuperSandro2000 (commented):
How do we want to include global rules in here for obviously nonsense CVEs?
Do we want to add a notice if rules no longer match anything?
We only want to use the keywords, not severity scores?

docs/api-spec.md
majewsky: The former, as explained in the table below.

docs/api-spec.md
majewsky: Yes.

docs/api-spec.md
majewsky: Well.... both? The manifest is the physical representation of what appears to the user as an image. I've changed the phrasing to "the respective image manifest".

docs/api-spec.md
majewsky: This can be added later if the need arises. For now, I'll confine myself to the feature set that we discussed with the Heureka team.

docs/api-spec.md
```diff
@@ -351,6 +327,72 @@ token needs to be presented when creating a replica of this primary account.
 Sublease tokens can only be issued for primary accounts. If the account in question is a replica account, 400 (Bad
 Request) is returned.
 
+## GET /keppel/v1/accounts/:name/trivy\_policies
```
SuperSandro2000, line 330: Why did we name this trivy_polices and not security_policies or similar? It is not really deeply tied into trivy.

internal/api/keppel/accounts.go
```diff
@@ -1013,3 +1015,129 @@ func (a *API) handlePostAccountSublease(w http.ResponseWriter, r *http.Request)
...
+		string(jsonBuf), account.Name)
+	if respondwith.ErrorText(w, err) {
+		return
+	}
+
+	//generate audit events
+	submitAudit := func(action cadf.Action, target audittools.TargetRenderer) {
+		if userInfo := authz.UserIdentity.UserInfo(); userInfo != nil {
+			a.auditor.Record(audittools.EventParameters{
+				Time:       time.Now(),
```
SuperSandro2000, line 1115: Should we use the mocked time now here?

internal/keppel/trivy_policy.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	if t.VulnerabilityIDRx == "" {
+		errs.Addf(`%s must have the "match_vulnerability_id" attribute`, path)
+	}
+
+	if t.Action.Assessment == "" {
+		errs.Addf(`%s.action must have the "assessment" attribute`, path)
+	}
+	if len(t.Action.Assessment) > 1024 {
+		errs.Addf(`%s.action.assessment cannot have a value larger than 1 KiB`, path)
```
SuperSandro2000, line 82:
```suggestion
		errs.Addf(`%s.action.assessment cannot be a larger than 1 KiB`, path)
```

docs/api-spec.md
majewsky: I agree on concern of tying the name too closely to the backing service. However, "security policy" is too vague in my opinion. I'll change it to "security scan policy".

internal/api/keppel/accounts.go
majewsky: This follows the existing implementation of audit events in other API handlers. Audit events need more than just time mocked out (e.g. each event also gets a v4 UUID for identification within Hermes). Therefore, we have a [normalization pass](https://github.com/sapcc/keppel/blob/3cb0cf024964539af63994e0481c4b4dcde376b7/internal/test/mock_auditor.go#L81) in the mock auditor that takes care of removing this real time value.

internal/keppel/security_scan_policy.go
```diff
@@ -110,3 +112,76 @@ func isSeverityKnownByTrivy(severity clair.VulnerabilityStatus) bool {
...
+	if p.NegativeRepositoryRx != "" && p.NegativeRepositoryRx.MatchString(repo.Name) {
+		return false
+	}
+	return p.RepositoryRx.MatchString(repo.Name)
+}
+
+// MatchesVulnerability evaluates the vulnerability regexes and checkin this policy.
+func (p SecurityScanPolicy) MatchesVulnerability(vuln trivy.ReportedVulnerability) bool {
+	fixIsReleased := vuln.FixedVersion != ""
+	if p.ExceptFixReleased && fixIsReleased {
```
SuperSandro2000, line 138:
```suggestion
	if p.ExceptFixReleased && vuln.FixedVersion != "" {

```

internal/tasks/manifests_test.go
```diff
@@ -755,3 +757,136 @@ func TestCheckVulnerabilitiesForNextManifestWithError(t *testing.T) {
 		`, image.Manifest.Digest, s.Clock.Now().Add(60*time.Minute).Unix(), s.Clock.Now().Unix(), clair.LowSeverity)
 	})
 }
+
+func TestCheckTrivySecurityStatusWithPolicies(t *testing.T) {
+	test.WithRoundTripper(func(_ *test.RoundTripper) {
+		j, s := setup(t, test.WithClairDouble, test.WithTrivyDouble)
```
SuperSandro2000, line 764:
```suggestion
		j, s := setup(t, test.WithTrivyDouble)
```

internal/tasks/manifests_test.go
```diff
@@ -755,3 +757,136 @@ func TestCheckVulnerabilitiesForNextManifestWithError(t *testing.T) {
...
+		//test baseline without policies
+		s.Clock.StepBy(1 * time.Hour)
+		expectSuccess(t, trivyJob.ProcessOne(s.Ctx))
+		expectError(t, sql.ErrNoRows.Error(), trivyJob.ProcessOne(s.Ctx))
+		tr.DBChanges().AssertEqualf(`
+			UPDATE blobs SET blocks_vuln_scanning = FALSE WHERE id = 1 AND account_name = 'test1' AND digest = '%[1]s';
+			UPDATE trivy_security_info SET vuln_status = '%[2]s', next_check_at = %[3]d, checked_at = %[4]d, check_duration_secs = 0 WHERE repo_id = 1 AND digest = '%[5]s';
+		`,
+			image.Layers[0].Digest,
+			clair.CriticalSeverity, s.Clock.Now().Add(60*time.Minute).Unix(), s.Clock.Now().Unix(), image.Manifest.Digest)
```
SuperSandro2000, line 783: I really don't like that go fmt is not strict about this new line here

internal/tasks/manifests_test.go
```diff
@@ -756,3 +758,136 @@ func TestCheckVulnerabilitiesForNextManifestWithError(t *testing.T) {
...
+
+			s.Clock.StepBy(1 * time.Hour)
+			expectSuccess(t, trivyJob.ProcessOne(s.Ctx))
+			expectError(t, sql.ErrNoRows.Error(), trivyJob.ProcessOne(s.Ctx))
+
+			tr.DBChanges().AssertEqualf(`
+				UPDATE trivy_security_info SET vuln_status = '%[1]s', next_check_at = %[2]d, checked_at = %[3]d WHERE repo_id = 1 AND digest = '%[4]s';
+			`, severity, s.Clock.Now().Add(60*time.Minute).Unix(), s.Clock.Now().Unix(), image.Manifest.Digest)
+		}
+
```
SuperSandro2000, line 803: I think we are missing tests for the constraints here, aka where the API would return an error.

internal/keppel/security_scan_policy.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// String returns the JSON representation of this policy (for use in log and
+// error messages).
+func (p SecurityScanPolicy) String() string {
+	//we only obtain SecurityScanPolicy instances through unmarshaling, so it is
+	//safe to assume that they will marshal without error
+	buf, err := json.Marshal(p)
+	if err != nil {
+		panic(err.Error())
```
SuperSandro2000, line 60: Shouldn't we return some contex here? Otherwise the error message would be hard to track down.

internal/tasks/manifests_test.go
majewsky: I was trying to avoid overlong lines, and matched the order of formatting directives.

internal/keppel/security_scan_policy.go
majewsky: Done.

internal/keppel/security_scan_policy.go
SuperSandro2000: Probably something about the securityScanPolicy.

If the database entry is somehow bit rotten, we could get into an error here.

majewsky: That would fail on **unmarshal**, not on marshal.

PR #234, opened 2023-06-01
Convert all old jobs to jobloop library except clair

SuperSandro2000:


internal/tasks/manifests.go
```diff
@@ -164,31 +151,27 @@ var syncManifestCleanupEmptyQuery = sqlext.SimplifyWhitespace(`
...
+			return repo, err
+		},
+		ProcessTask: j.processSyncManifests,
+	}).Setup(registerer)
+}
 
+func (j *Janitor) processSyncManifests(_ context.Context, repo keppel.Repository, _ prometheus.Labels) error {
 	//find repository to sync
 	err := j.db.SelectOne(&repo, syncManifestRepoSelectQuery, j.timeNow())
 	if err != nil {
```
majewsky, line 174: The ProcessTask function should not repeat the discovery query.

majewsky (changes requested):

internal/tasks/manifests.go
```diff
@@ -64,40 +64,31 @@ var outdatedManifestSearchQuery = sqlext.SimplifyWhitespace(`
...
-		return err
-	}
+// than 6 hours. At most one manifest is validated per call.
+func (j *Janitor) ValidateManifestJob(registerer prometheus.Registerer) jobloop.Job {
+	return (&jobloop.ProducerConsumerJob[keppel.Manifest]{
+		Metadata: jobloop.JobMetadata{
+			ReadableName: "manifest syncs in replica repos",
+			CounterOpts: prometheus.CounterOpts{
+				Name: "keppel_blob_validations",
+				Help: "Counter for blob validations.",
```
majewsky, line 74: There's a copy-paste error here, which now makes me concerned there's going to be more copy-paste errors elsewhere. Please have another look.

majewsky (changes requested):

majewsky (approved):
I pushed one more commit to clean up the naming of functions and the references to those names throughout docstrings and comments. If this looks good to you, please go ahead with the merge.

SuperSandro2000:
I was so focused on converting to the new library that I totally overlooked that those names are not that great. Thanks for cleaning up!

PR #239, opened 2023-06-07
add X-Keppel-Applicable-Policies field to Trivy reports

majewsky:


internal/keppel/security_scan_policy.go
```diff
@@ -184,3 +185,58 @@ func (s SecurityScanPolicySet) PolicyForVulnerability(vuln trivy.ReportedVulnera
...
+	//
+	//    {...} -> {...,"X-Keppel-Applicable-Policies":{...}}
+	applicablePoliciesJSON, err := json.Marshal(applicablePolicies)
+	if err != nil {
+		return fmt.Errorf("cannot marshal X-Keppel-Applicable-Policies: %w", err)
+	}
+	payload.Contents = bytes.Join([][]byte{
+		//this trim cannot fail because we already did a successful json.Unmarshal above
+		bytes.TrimSuffix(bytes.TrimSpace(payload.Contents), []byte("}")),
+		[]byte(`,"X-Keppel-Applicable-Policies":`),
```
SuperSandro2000, line 236:
```suggestion
		bytes.TrimSuffix(bytes.TrimSpace(payload.Contents), []byte("\n}")),
		[]byte(`,\n"X-Keppel-Applicable-Policies":`),
```
I think this would generate nicer looking json but it is also more brittle. I am not sure if the trade off is worth it.

majewsky: I changed the code to use the upstream types, since that does not add too many new dependencies (only a few).

This removes the manual JSON mangling shenanigans and introduces a different set of shenanigans for enriching the upstream types with our custom fields. I do think that this new set of shenanigans should be less brittle though.

internal/keppel/security_scan_policy.go
```diff
@@ -184,3 +185,58 @@ func (s SecurityScanPolicySet) PolicyForVulnerability(vuln trivy.ReportedVulnera
...
+		applicablePoliciesJSON,
+		[]byte("}\n"),
```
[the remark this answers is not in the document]
majewsky: Yeah, there was an extra newline in the fixture. I removed it.

SuperSandro2000 (approved):

PR #250, opened 2023-07-04
Add go routine parallelism to trivy jobloop

SuperSandro2000:


internal/tasks/manifests.go
```diff
@@ -894,39 +900,84 @@ func (j *Janitor) CheckTrivySecurityStatusJob(registerer prometheus.Registerer)
...
-		DiscoverRow: func(_ context.Context, tx *gorp.Transaction, _ prometheus.Labels) (securityInfo keppel.TrivySecurityInfo, err error) {
-			err = tx.SelectOne(&securityInfo, securityCheckSelectQuery, j.timeNow())
-			return securityInfo, err
+		DiscoverRow: func(_ context.Context, tx *gorp.Transaction, _ prometheus.Labels) (securityInfos []keppel.TrivySecurityInfo, err error) {
+			_, err = tx.Select(&securityInfos, securityCheckSelectQuery, j.timeNow())
+
+			// TODO: How to convince tx.Select to do this?
+			if len(securityInfos) == 0 {
+				err = sql.ErrNoRows
+			}
```
SuperSandro2000, line 909: @majewsky what am I doing wrong here?

internal/tasks/manifests.go
```diff
@@ -894,39 +900,84 @@ func (j *Janitor) CheckTrivySecurityStatusJob(registerer prometheus.Registerer)
...
-	}
-	manifest, err := keppel.FindManifest(tx, *repo, securityInfo.Digest)
-	if err != nil {
-		return fmt.Errorf("cannot find manifest for repo %s and digest %s: %w", repo.FullName(), securityInfo.Digest, err)
+type returnStruct struct {
+	securityInfo keppel.TrivySecurityInfo
+	err          error
+}
+
+// processTrivySecurityInfo parallises the CheckTrivySecurityStatusJob jobloop without requiring extra database connections
```
majewsky, line 922: Nitpick:
```suggestion
// processTrivySecurityInfo parallelises the CheckTrivySecurityStatusJob jobloop without requiring extra database connections
```
"Paralyzing" is something very different. :)

internal/tasks/manifests.go
```diff
@@ -894,39 +900,84 @@ func (j *Janitor) CheckTrivySecurityStatusJob(registerer prometheus.Registerer)
...
+// It processes SecurityInfos in batches in the value of trivySecurityInfoBatchSize and runs the value of trivySecurityInfoThreads in parallel.
+func (j *Janitor) processTrivySecurityInfo(ctx context.Context, tx *gorp.Transaction, securityInfos []keppel.TrivySecurityInfo, labels prometheus.Labels) error {
+	// prevent deadlocks by waiting for more securityInfos in range below
+	lenSecurityInfo := len(securityInfos)
+	if trivySecurityInfoBatchSize < lenSecurityInfo {
+		trivySecurityInfoBatchSize = lenSecurityInfo
+	}
+
+	inputChan := make(chan keppel.TrivySecurityInfo, trivySecurityInfoBatchSize)
```
majewsky, line 931: Why is the batch size adjusted down here? Since it can only ever go down, it will tend to 0 over time. But since the SQL query still uses the original batch size, eventually the adjusted batch size will be too small and `inputChan` will not be able to hold all items.

internal/tasks/manifests.go
```diff
@@ -894,39 +900,84 @@ func (j *Janitor) CheckTrivySecurityStatusJob(registerer prometheus.Registerer)
...
+	for _, securityInfo := range securityInfos {
+		inputChan <- securityInfo
+	}
+	close(inputChan)
+
+	// start the valuf of trivySecurityInfoThreads channels
```
majewsky, line 937: Please proofread.

internal/tasks/manifests.go
```diff
@@ -894,39 +900,84 @@ func (j *Janitor) CheckTrivySecurityStatusJob(registerer prometheus.Registerer)
...
-		return fmt.Errorf("cannot find repo for manifest %s: %w", securityInfo.Digest, err)
-	}
-	account, err := keppel.FindAccount(tx, repo.AccountName)
-	if err != nil {
-		return fmt.Errorf("cannot find account for repo %s: %w", repo.FullName(), err)
-	}
-	manifest, err := keppel.FindManifest(tx, *repo, securityInfo.Digest)
-	if err != nil {
-		return fmt.Errorf("cannot find manifest for repo %s and digest %s: %w", repo.FullName(), securityInfo.Digest, err)
+type returnStruct struct {
```
majewsky, line 917: This type name is very vague. Since an accurate type name would be very convoluted, I recommend moving this type declaration inside the method that uses it. Then the name can be made more accurate without being overly long.

internal/tasks/manifests.go
```diff
@@ -1011,7 +1080,7 @@ func (j *Janitor) doSecurityCheck(ctx context.Context, account keppel.Account, r
 		parsedTrivyReport, err := j.cfg.Trivy.ScanManifestAndParse(ctx, tokenResp.Token, imageRef)
 		if err != nil {
 			logg.Error(err.Error())
-			return err
+			return keppel.TrivySecurityInfo{}, err
```
majewsky, line 1083: Returning a default-constructed value here breaks the `defer` logic.

internal/tasks/manifests.go
```diff
@@ -936,7 +987,24 @@ var securityInfoCheckSubmanifestInfoQuery = sqlext.SimplifyWhitespace(`
 		WHERE r.parent_digest = $1
 `)
 
-func (j *Janitor) doSecurityCheck(ctx context.Context, account keppel.Account, repo keppel.Repository, manifest keppel.Manifest, securityInfo *keppel.TrivySecurityInfo) (returnedError error) {
+func (j *Janitor) doSecurityCheck(ctx context.Context, inputSecurityInfo keppel.TrivySecurityInfo) (securityInfo keppel.TrivySecurityInfo, returnedError error) {
+	// we cannot use a pointer to the securityInfo because we want to return it into the returnChan
+	securityInfo = inputSecurityInfo
```
majewsky, line 992: Why though? The caller could just as well be rewritten from
```go
securityInfo, err := j.doSecurityCheck(ctx, inputSecurityInfo)
```
into
```go
err := j.doSecurityCheck(ctx, &securityInfo)
```
Then there is no risk of accidentally returning an empty `securityInfo` instead of the original one.

majewsky (changes requested):

internal/tasks/manifests.go
majewsky: Nothing. `Select` (or its std counterpart `sql.Conn.Query`) can legitimately yield 0 rows (as opposed to `SelectOne` or `sql.Conn.QueryRow`), so that is not considered an error. The check as written here is fine.

internal/tasks/manifests.go
SuperSandro2000: I converted this to a local variable.

PR #251, opened 2023-07-06
Bye, Bye clair 👋

SuperSandro2000:


majewsky (approved):
I have pushed two more commits to remove some other things that are not needed anymore, and adjust some documentation text and comments.

PR #269, opened 2023-08-02
Add context to all nework requests, enabel noctx linter

SuperSandro2000:


cmd/api/main.go
```diff
@@ -58,7 +58,7 @@ func AddCommandTo(parent *cobra.Command) {
 	parent.AddCommand(cmd)
 }
 
-func run(cmd *cobra.Command, args []string) {
+func run(_ *cobra.Command, _ []string) {
```
majewsky, line 61: I don't like that this loses information on what the arguments are about. For `*cobra.Command`, it's reasonably clear. But `[]string` could be lots of things. Does the linter allow keeping the names and assigning to blank:

```
func run(cmd *cobra.Command, args []string) {
  _, _ = cmd, args
```

That's the common idiom for marking that a variable is not used, similar to void-casting in C/C++.
If the linter allows that, I would prefer that for all places where the purpose of the arguments is not immediately clear from the typing.

internal/api/peer/delegatedpull.go
```diff
@@ -56,7 +56,8 @@ func (a *API) handleDelegatedPullManifest(w http.ResponseWriter, r *http.Request
 		Password: r.Header.Get("X-Keppel-Delegated-Pull-Password"), //may be empty
 	}
 	ref := models.ParseManifestReference(vars["reference"])
-	manifestBytes, manifestMediaType, err := rc.DownloadManifest(ref, &opts)
+	manifestBytes, manifestMediaType, err := rc.DownloadManifest(r.Context(), ref, &opts)
+	//nolint:errorlint // not applicable
```
majewsky, line 60: This description is incredibly vague. In Limes, I changed these to
```suggestion
	//nolint:errorlint // a type cast is clearer than errors.As()
```

Same for all the other identical nolint comments.

internal/drivers/openstack/keystone.go
```diff
@@ -149,6 +149,7 @@ func (d *keystoneDriver) AuthenticateUser(ctx context.Context, userName, passwor
 	)
 
 	if t.Err != nil {
+		//nolint:errorlint // the casted variable is used
```
majewsky, line 152: Another case where the nolint description can be clarified: `errors.As()` does support retrieving the casted error value, we just don't consider it worth the effort because it's more tedious than a straight-up type cast. Same for the other identical nolint descriptions below.

internal/keppel/errors.go
```diff
@@ -128,6 +128,7 @@ type RegistryV2Error struct {
 // AsRegistryV2Error tries to cast `err` into RegistryV2Error. If `err` is not a
 // RegistryV2Error, it gets wrapped in ErrUnknown instead.
 func AsRegistryV2Error(err error) *RegistryV2Error {
+	//nolint:errorlint // the casted variable is used
```
majewsky, line 131: This is a case where errors.As could maybe actually be used without making the implementation less clear.

majewsky (changes requested):
Side-note: Since our main objection to `errors.As()` is its clunky interface, I will take a look whether we can wrap that in a type-safe interface using generics.

PR #276, opened 2023-08-11
Base rate limit on source ip

SuperSandro2000:


internal/api/utils.go
```diff
@@ -41,7 +41,7 @@ func CheckRateLimit(r *http.Request, rle *keppel.RateLimitEngine, account keppel
 		return nil
 	}
 
-	allowed, result, err := rle.RateLimitAllows(r.Context(), account, action, amount)
+	allowed, result, err := rle.RateLimitAllows(r.Context(), r.RemoteAddr, account, action, amount)
```
majewsky, line 44: Please use `httpext.RequesterIPFor(r)` instead or `r.RemoteAddr`. In most real-world cases, we are behind reverse proxies and need to inspect the `X-Forwarded-For` header instead of the current source IP of the request.

majewsky (changes requested):

majewsky (approved):

PR #278, opened 2023-08-28
Don't treat a 404 as an error

SuperSandro2000:


majewsky (changes requested):
1. `schwift.Is()` is more specific and idiomatic here than `errext.As`.
2. Errors should be ignored earlier. As it is right now, a 404 followed by a real error would be reported as success.

I put a counter-proposal in #281.

PR #280, opened 2023-09-04
Add test for trivy ratelimit

SuperSandro2000:
@majewsky somehow the rate limit is not triggered and I couldn't find the problem yet. Some help is appreciated.

### DIFF
--- a/internal/api/keppel/manifests_test.go
+++ b/internal/api/keppel/manifests_test.go
@@ -20,19 +20,25 @@ package keppelv1_test
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"sort"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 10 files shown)


majewsky (changes requested):
The test is failing because you did not provide the `test.WithKeppelAPI` setup option. The `404 not found` is rendered by the router because the respective route does not exist.

This is pointing to a more general concern: Since this test is covering an endpoint in the Keppel API, it should be moved into `internal/api/keppel/`.

majewsky (changes requested):
Why are you moving all these test functions from `internal/api/registry/` to `internal/test/`? They are all very specific to registry API tests. The Trivy ratelimit test does not care about the anycast option, which is the main source of complexity for these test helpers. The Trivy rate limit test only needs the basic `test.NewSetup()` with a fixed list of options and nothing else.

majewsky (approved):

PR #296, opened 2023-10-13
Improve error messages while syncing tags from mirrors 

SuperSandro2000:
…eneral


majewsky:
I have bumped Schwift to 1.3.0. With this update, most of this PR should be obsolete.

PR #301, opened 2023-10-23
Change rate limit datatype to uint32 to not lose precision when casting

SuperSandro2000:


internal/api/registry/blobs.go
```diff
@@ -118,7 +118,7 @@ func (a *API) handleGetOrHeadBlob(w http.ResponseWriter, r *http.Request) {
 		//AnycastBlobBytePullAction is only relevant for GET requests since it
 		//limits the size of the response body (which is empty for HEAD)
 		if r.Method == http.MethodGet {
-			err = api.CheckRateLimit(r, a.rle, *account, authz, keppel.AnycastBlobBytePullAction, blob.SizeBytes)
+			err = api.CheckRateLimit(r, a.rle, *account, authz, keppel.AnycastBlobBytePullAction, int(blob.SizeBytes))
```
SuperSandro2000, line 121: Either way we hider here that the rate limiter caps at 2.147 GB or max int32 aka 2147483647 bytes.

SuperSandro2000:
As discussed internally, changing the type to uint32 to not lose the upper half when calculating for big images.

majewsky:
Megabyte precision for blob pulls is too bad for me. A significant number of blobs is in the KiB range or below, e.g. image config blobs or small `RUN` layers that only change a config file. Rounding these up to megabytes is an unacceptable loss in precision compared to the rather academic problem of overflowing on an uint64 -> int64 conversion. I have added an explicit overlfow check in a0d81437d5 that just rejects blob pulls larger than 2^63 bytes (about 8 million TiB).

PR #302, opened 2023-10-23
Add context to all database calls

SuperSandro2000:
Also fix all unused parameters

I wanted to achieve a maintainable solution that also works for future addition. For that reason I choose to mirror the interface in internal/keppel/database.go and mark all the functions deprecated which then get marked ci golangci-lint.

### DIFF
--- a/cmd/api/gui_redirect.go
+++ b/cmd/api/gui_redirect.go
@@ -69,7 +69,7 @@ func (g *guiRedirecter) tryRedirectToGUI(w http.ResponseWriter, r *http.Request)
 
 	//is it publicly readable?
 	var policies []keppel.RBACPolicy
-	_, err = g.db.Select(&policies,
+	_, err = g.db.WithContext(r.Context()).Select(&policies,
 		"SELECT * FROM rbac_policies WHERE can_anon_pull AND account_name = $1",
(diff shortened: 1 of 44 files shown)


majewsky:
I saw several places where WithContext is not used, especially around transactions. The problem is that `db.WithContext(ctx)` does not have a Begin method. Apparently Gorp expects one to do Begin without context and then use `tx.WithContext(ctx)` after the transaction has already started.

Since there are some places where WithContext is not used, I tried to have an interface that more strongly discourages misuse. For example, in `internal/api/keppel/api.go`,

```go
type API struct {
  ...
  db *keppel.DB
  ...
}
```

could be changed to

```go
type API struct {
  ...
  dbWithoutContext *keppel.DB // do not use directly
  ...
}

func (a*API) db(ctx context.Context) gorp.SqlExecutor {
  return a.db.WithContext(ctx)
}
```

to make it clear that the DB handle should only be accessed through the helper method that needs a context. But then we have a `gorp.SqlExecutor` which is missing some vital methods, specifically `Begin` and `Prepare` (and therefore also cannot be cast into `sqlext.Executor` which several library functions take). This looks like a significant issue for context coverage. Opinions?

PR #351, opened 2024-02-28
replace `rbac_policies` table with `accounts.rbac_policies_json`

majewsky:
This simplifies a lot of the CRUD for accounts and RBAC policies, and reduces the number of SQL queries for each auth by one, at the expense of making `accounts` records a bit larger.

I have verified that the logic from the previous PR has migrated all existing policies into the new DB column, so it is safe to drop the old table.

### DIFF
--- a/internal/keppel/rbac_policy.go
+++ b/internal/keppel/rbac_policy.go
@@ -20,16 +20,113 @@
 package keppel
 
 import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net"
+
 	"github.com/sapcc/go-bits/regexpext"
 )
 
-// NewRBACPolicy is a policy granting user-defined access to repos in an account.
+// RBACPolicy is a policy granting user-defined access to repos in an account.
 // It is stored in serialized form in the RBACPoliciesJSON field of type Account.
-//
-// TODO: rename to type RBACPolicy after the `rbac_policies` table has been removed
-type NewRBACPolicy struct {
-	CidrPattern       string                `json:"match_cidr,omitempty"`
-	RepositoryPattern regexpext.PlainRegexp `json:"match_repository,omitempty"`
-	UserNamePattern   regexpext.PlainRegexp `json:"match_username,omitempty"`
-	Permissions       []string              `json:"permissions"`
+type RBACPolicy struct {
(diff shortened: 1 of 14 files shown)


internal/keppel/rbac_policy.go
```diff
@@ -20,16 +20,113 @@
...
-	CidrPattern       string                `json:"match_cidr,omitempty"`
-	RepositoryPattern regexpext.PlainRegexp `json:"match_repository,omitempty"`
-	UserNamePattern   regexpext.PlainRegexp `json:"match_username,omitempty"`
-	Permissions       []string              `json:"permissions"`
+type RBACPolicy struct {
+	CidrPattern       string                  `json:"match_cidr,omitempty"`
+	RepositoryPattern regexpext.BoundedRegexp `json:"match_repository,omitempty"`
+	UserNamePattern   regexpext.BoundedRegexp `json:"match_username,omitempty"`
+	Permissions       []RBACPermission        `json:"permissions"`
+}
```
majewsky, line 38: FYI: The use of `PlainRegexp` was copied from `internal/api/keppel/accounts.go`, but turned out to be incorrect because the match patterns are defined as implying `^` and `$` anchors, i.e. the BoundedRegexp behavior. This was not a problem before because the old type in the API package only did input validation, not actual matching.

internal/keppel/rbac_policy.go
```diff
@@ -20,16 +20,113 @@
...
+// RBACPermission enumerates permissions that can be granted by an RBAC policy.
+type RBACPermission string
+
+const (
+	PolicyAllowsPull               RBACPermission = "pull"
+	PolicyAllowsPush               RBACPermission = "push"
+	PolicyAllowsDelete             RBACPermission = "delete"
+	PolicyAllowsAnonymousPull      RBACPermission = "anonymous_pull"
+	PolicyAllowsAnonymousFirstPull RBACPermission = "anonymous_first_pull"
+)
```
majewsky, line 49: FYI: I'm not entirely happy with the names for these enum variants, but it's the best I could come up with. Anything that contains "RBAC" ends up looking like a keysmash, e.g. `AnonymousFirstPullRBACPermission`.

SuperSandro2000: Maybe we can drop the Allows and just take that for implicit

PR #356, opened 2024-03-14
Add AccountManagementDriver

SuperSandro2000:
TODO: 
- [ ] add a test which modifies a managed account
- [ ] add a test which adds, modifies and deletes a federated account


internal/api/keppel/accounts.go
```diff
@@ -62,11 +62,11 @@ type Account struct {
...
 type ReplicationPolicy struct {
-	Strategy string
+	Strategy string `json:"strategy"`
 	//only for `on_first_use`
-	UpstreamPeerHostName string
+	UpstreamPeerHostName string `json:"upstream_peer_hostname"`
 	//only for `from_external_on_first_use`
-	ExternalPeer ReplicationExternalPeerSpec
+	ExternalPeer ReplicationExternalPeerSpec `json:"external_peer"`
 }
```
majewsky, line 70: For a PoC, this is fine. For the final implementation, this should move into `package keppel`. In general, to keep the dependency graph orderly, my rule of thumb is to put everything that drivers need into `package keppel`, and everything that APIs and janitor tasks need into one of `package keppel/processor/auth`. There should not be any horizontal dependencies between APIs, janitor tasks and drivers.

The existing Policy types in `package keppel` can serve as a guideline to inform the kind of interface offered to the API and driver implementation.

internal/drivers/basic/account_management_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"github.com/sapcc/go-bits/assert"
+)
+
+func TestAccountManagementDriver(t *testing.T) {
+	driver := AccountManagementDriver{
+		skipConfigLoading: true,
+	}
+
+	driver.config.Accounts = []Accounts{
```
majewsky, line 36: I would find it nicer to exercise more of the driver code by loading this from a file in a `fixtures` directory.

majewsky (commented):
Good start.

internal/processor/accounts.go
```diff
@@ -64,7 +64,10 @@ func (p *Processor) GetPlatformFilterFromPrimaryAccount(ctx context.Context, pee
 var looksLikeAPIVersionRx = regexp.MustCompile(`^v[0-9][1-9]*$`)
 
 // CreateOrUpdate can be used on an API account and returns the database representation of it.
-func (p *Processor) CreateOrUpdateAccount(ctx context.Context, account keppel.Account, fd keppel.FederationDriver, userInfo audittools.UserInfo, r *http.Request, getSubleaseToken func(models.Peer) (string, *keppel.RegistryV2Error)) (models.Account, *keppel.RegistryV2Error) {
+func (p *Processor) CreateOrUpdateAccount(ctx context.Context, account keppel.FullAccount, fd keppel.FederationDriver, userInfo audittools.UserInfo, r *http.Request, getSubleaseToken func(models.Peer) (string, *keppel.RegistryV2Error)) (models.Account, *keppel.RegistryV2Error) {
```
majewsky, line 67: The `fd` argument is not required anymore since Processor gains a FederationDriver slot in this PR.

majewsky (changes requested):

internal/drivers/basic/account_management.go
```diff
@@ -0,0 +1,129 @@
...
+
+	decoder := yaml.NewDecoder(reader)
+	decoder.SetStrict(true)
+	var config AccountConfig
+	err = decoder.Decode(&config)
+	if err != nil {
+		return err
+	}
+
+	a.config = config
```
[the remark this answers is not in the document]
SuperSandro2000: We load the config file each time we discover a new task which should be good enough.

internal/drivers/basic/account_management.go
```diff
@@ -0,0 +1,137 @@
...
+		return &AccountManagementDriver{}
+	})
+}
+
+// PluginTypeID implements the keppel.AccountManagementDriver interface.
+func (a *AccountManagementDriver) PluginTypeID() string { return "basic" }
+
+// Init implements the keppel.AccountManagementDriver interface.
+func (a *AccountManagementDriver) Init() error {
+	configPath, err := osext.NeedGetenv("KEPPEL_ACCOUNT_MANAGEMENT_FILE")
```
majewsky, line 65: The common naming convention would be:
```suggestion
	configPath, err := osext.NeedGetenv("KEPPEL_ACCOUNT_MANAGEMENT_CONFIG_PATH")
```

internal/drivers/basic/account_management.go
```diff
@@ -0,0 +1,137 @@
+/******************************************************************************
```
majewsky, line 1: Documentation for this driver should be added to `docs/drivers`. Please take a look at the existing driver docs for reference.

internal/drivers/basic/account_management.go
```diff
@@ -0,0 +1,137 @@
...
+	configPath, err := osext.NeedGetenv("KEPPEL_ACCOUNT_MANAGEMENT_FILE")
+	if err != nil {
+		return err
+	}
+	a.ConfigPath = configPath
+	return a.loadConfig()
+}
+
+// ConfigureAccount implements the keppel.AccountManagementDriver interface.
+func (a *AccountManagementDriver) ConfigureAccount(modelAccount models.Account) (*keppel.Account, []keppel.SecurityScanPolicy, error) {
```
majewsky, line 74: Sorry for only noticing this on the third or fourth round of reviews, and also not noticing this when I specified this interface in the ticket, but this should rather take an `accountName string` argument instead of the entire account model. Giving an entire DB row to a driver method invites way stronger coupling than we should have here.

internal/drivers/basic/account_management_test.go
```diff
@@ -0,0 +1,113 @@
...
+		Name:         "abcde",
+		AuthTenantID: "1245",
+		GCPolicies: []keppel.GCPolicy{
+			{
+				Action:               "protect",
+				NegativeRepositoryRx: "archive/.*",
+				RepositoryRx:         ".*/database",
+				TimeConstraint: &keppel.GCTimeConstraint{
+					FieldName: "pushed_at",
+					MaxAge:    21600000000000,
```
majewsky, line 61:
```suggestion
					MaxAge:    6 * time.Hour,
```

internal/drivers/basic/fixtures/account_management.yaml
```diff
@@ -0,0 +1,38 @@
+accounts:
+- name: abcde
+  auth_tenant_id: "1245"
+  gc_policies:
+  - match_repository: .*/database
+    except_repository: archive/.*
+    time_constraint:
+      "on": pushed_at
+      newer_than: 21600000000000
```
majewsky, line 9: Since this file is supposed to be human-writable, this should use a more accessible format. We have a precedent for this in https://pkg.go.dev/github.com/sapcc/limes/internal/util#MarshalableTimeDuration, so this would make sense for go-bits (maybe a new subpackage "timeext").

internal/tasks/account_management.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/sapcc/go-bits/logg"
+
+	"github.com/sapcc/keppel/internal/auth"
+	peerclient "github.com/sapcc/keppel/internal/client/peer"
+	"github.com/sapcc/keppel/internal/keppel"
+	"github.com/sapcc/keppel/internal/models"
+)
+
+// EnforceManagedAccounts is a job. Each task creates newly discovered accounts from the driver.
+func (j *Janitor) EnforceManagedAccounts(registerer prometheus.Registerer) jobloop.Job {
```
majewsky, line 40: Precedent for functions returning a `Job` is to have their name end in `Job`:
```suggestion
func (j *Janitor) EnforceManagedAccountsJob(registerer prometheus.Registerer) jobloop.Job {
```

internal/tasks/account_management.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return nil, err
+	}
+	accountNames = append(accountNames, toEnforceAccountNames...)
+
+	if len(accountNames) == 0 {
+		return nil, sql.ErrNoRows
+	}
+
+	return accountNames, nil
```
majewsky, line 83: I typed out several review comments that I ended up deleting when I gradually understood more about what this function is doing. This definitely needs some "coding in paragraphs" level comments to outline the overall approach.

Also, I would suggest returning only a single account name at a time to help clarify the approach:

- If there are ManagedAccountNames that do not have a DB record with `is_managed`, return the first such name.
- Otherwise, return the next account name that has `next_account_enforcement_at < NOW()` (in chronological order).

internal/tasks/account_management.go
```diff
@@ -0,0 +1,166 @@
...
+			accountModel.InMaintenance = true
+
+			resp, err := j.processor().DeleteAccount(ctx, *accountModel)
+			if err != nil {
+				return err
+			}
+			if resp != nil {
+				logg.Error("Deleting account %s failed: %s", accountName, resp.Error)
+				return errors.New(resp.Error)
+			}
```
majewsky, line 117: I have thought about how we can best deal with the problem that account deletion is not a one-step thing, and I have a way that should work.

- When `resp` indicates to the caller that they need to do work (e.g. deleting manifests), do it right here, then recurse back into another try of `DeleteAccount`. The entire DeleteAccount thing should probably be moved into its own function to allow for easy recursion by tail call.
- When `resp` indicates to the caller that they need to wait for background tasks to happen (the whole blob deletion thing), update `NextAccountManagementAt` with a small delay (e.g. 1 minute instead of 1 hour) to come back to this account soon.

internal/tasks/account_management.go
```diff
@@ -0,0 +1,166 @@
...
+			_, rerr := j.processor().CreateOrUpdateAccount(ctx, *account, userIdentity.UserInfo(), janitorDummyRequest, getSubleaseToken, setCustomFields)
+			if rerr != nil {
+				return rerr
+			}
+		}
+
+		_, err = j.db.Exec(accountAnnouncementDoneQuery, accountName, j.timeNow().Add(j.addJitter(1*time.Hour)))
+		if err != nil {
+			return err
+		}
```
majewsky, line 162: This is the wrong query. You're updating `NextFederationAnnouncementAt` instead of `NextAccountEnforcementAt` here. I assume this is a copy-paste error. If so, please do a sweep if there are other copy-paste errors like this that I did not catch.

internal/tasks/account_management_test.go
```diff
@@ -0,0 +1,69 @@
...
+	`)
+
+	// test if errors are propagated
+	s.AMD.ConfigPath = "./fixtures/account_management_error.yaml"
+	s.Clock.StepBy(2 * time.Hour)
+	expectError(t, processor.ErrAccountNameEmpty.Error(), createManagedAccountsJob.ProcessOne(s.Ctx))
+	tr.DBChanges().AssertEmpty()
+
+	// and delete the account again
```
majewsky, line 61: There should be another test case where a managed account is created and deleted, but inbetween an image is uploaded into it. Running this test to completion will probably require driving some other GC jobs in parallel to fully execute the account deletion.

internal/tasks/fixtures/account_management_error.yaml
```diff
@@ -0,0 +1,2 @@
+accounts:
+- name: ""
```
majewsky, line 2: I don't see this file being used in the test. Probably a leftover before this test got moved to `internal/drivers/basic`?

majewsky (changes requested):
I regret not finding a way to split this PR further into more digestible chunks. I feel like in every round, I'm looking with a different focus and thus seeing lots of new stuff that I could have caught earlier.

docs/drivers/account-management-basic.md
```diff
@@ -0,0 +1,25 @@
...
+
+The configuration file must be formatted in YAML.
+The following fields are valid inside the configuration file:
+
+| Field | Explanation |
+| ----- | ----------- |
+| `accounts` | list of objects | A list of objects, one for each managed account. Any managed accounts that exists in the database, but is not included in this list will be deleted. |
+| `accounts[].name`<br>`accounts[].auth_tenant_id`<br>`accounts[].gc_policies`<br>`accounts[].rbac_policies`<br>`accounts[].replication`<br>`accounts[].validation` | These fields have the same structure and meaning as on `{GET,PUT} /keppel/v1/accounts/:name`; see [API spec](../api-spec.md) for details. |
+| `accounts[].security_scan_policies` | This field has the same structure and meaning as `policies` on `{GET,PUT}
+/keppel/v1/accounts/:name/security_scan_policies`; see [API spec](../api-spec.md) for details. |
```
SuperSandro2000, line 25:
```suggestion
| `accounts[].security_scan_policies` | This field has the same structure and meaning as `policies` on
`{GET,PUT}/keppel/v1/accounts/:name/security_scan_policies`; see [API spec](../api-spec.md) for details. |
```

majewsky (changes requested):
Note to self: The API should reject PUT on accounts with `IsManaged = true`.

internal/tasks/account_management_test.go
```diff
@@ -0,0 +1,172 @@
...
+	tr.DBChanges().AssertEqualf(`
+		UPDATE accounts SET in_maintenance = TRUE, next_enforcement_at = %[1]d WHERE name = 'abcde';
+		DELETE FROM manifest_blob_refs WHERE repo_id = 2 AND digest = '%[2]s' AND blob_id = 1;
+		DELETE FROM manifest_blob_refs WHERE repo_id = 2 AND digest = '%[2]s' AND blob_id = 2;
+		DELETE FROM manifest_blob_refs WHERE repo_id = 2 AND digest = '%[2]s' AND blob_id = 3;
+		DELETE FROM manifest_contents WHERE repo_id = 2 AND digest = '%[2]s';
+		DELETE FROM manifests WHERE repo_id = 2 AND digest = '%[2]s';
+		DELETE FROM tags WHERE repo_id = 2 AND name = 'latest';
+		DELETE FROM trivy_security_info WHERE repo_id = 2 AND digest = '%[2]s';
+		`,
```
SuperSandro2000, line 120: gofmt sadly doesn't understand this

majewsky: Based on the indentation in existing instances of AssertEqualf, I have made a guess regarding what you mean by "this". Please take a look.

PR #393, opened 2024-06-13
bump trivy to 0.52.1

majewsky:
This was way more complicated than previous updates because Trivy moved their package structure around. Refering to anything from `package types` now pulls in basically the entirety of Trivy's humongous dependency graph, which we want to avoid. To avoid this, we use

- `type Results` from the field-compatible declaration in the WASM-compatible section of Trivy
- `type Report` and `type Metadata` from local copies of these types with fields replaced by generic types where necessary

I also had to delete the unit test that checks that our types have the same fields as `types.Report`, because once again I cannot refer to that type without blowing up the size of the vendor directory. Instead, we now deserialize reports in strict mode to notice new fields being added in Trivy through errors in QA (strict mode parsing will make Keppel fail when it encounters unexpected fields in reports from Trivy).


internal/trivy/report.go
```diff
@@ -0,0 +1,53 @@
...
+*
+*******************************************************************************/
+
+package trivy
+
+import (
+	"time"
+
+	ftypes "github.com/aquasecurity/trivy/pkg/fanal/types"
+	"github.com/aquasecurity/trivy/pkg/module/serialize"
```
SuperSandro2000, line 26:
```suggestion
	stypes "github.com/aquasecurity/trivy/pkg/module/serialize"
```

SuperSandro2000 (approved):

PR #398, opened 2024-07-01
Upgrade to gophercloud 2.0.0

SuperSandro2000:
Doesn't compile yet, as I haven't found a way to extract the Retry-After header in internal/drivers/openstack/keystone.go


internal/drivers/openstack/keystone.go
```diff
@@ -126,23 +126,20 @@ func (d *keystoneDriver) AuthenticateUser(ctx context.Context, userName, passwor
...
 
 	t := d.TokenValidator.CheckCredentials(
 		fmt.Sprintf("username=%s,password=%s", userName, password),
-		func() gopherpolicy.TokenResult { return tokens.Create(&throwAwayClient, &authOpts) },
+		func() gopherpolicy.TokenResult { return tokens.Create(ctx, &throwAwayClient, &authOpts) },
 	)
 
 	if t.Err != nil {
-		if err, ok := errext.As[gophercloud.ErrDefault429](t.Err); ok {
+		if err, ok := errext.As[gophercloud.ErrUnexpectedResponseCode](t.Err); ok {
```
majewsky, line 142: Should also check that response code is 429.

internal/tasks/uploads_test.go
```diff
@@ -52,7 +53,7 @@ func TestDeleteAbandonedUploadWithZeroChunks(t *testing.T) {
 func TestDeleteAbandonedUploadWithOneChunk(t *testing.T) {
 	testDeleteUpload(t, func(sd keppel.StorageDriver, account models.Account) models.Upload {
 		data := "just some test data"
-		err := sd.AppendToBlob(account, testStorageID, 1, p2len(data), strings.NewReader(data))
+		err := sd.AppendToBlob(context.TODO(), account, testStorageID, 1, p2len(data), strings.NewReader(data))
```
majewsky, line 56: I would like to see `s.Ctx` being used here. Please amend `testDeleteUpload` to provide the `s test.Setup` variable to its callback.

majewsky (changes requested):

majewsky (approved):

PR #406, opened 2024-07-24
add LIQUID support

majewsky:
There is already a quota/usage API implementation in `internal/api/keppel/` that does roughly the same thing, so this mostly just translates to and from the payload types used by LIQUID. The diff ended up being a bit bigger because I moved some stuff around to avoid code duplication.


VoigtS (approved):
I like the split of responsibilities in the in the api definitions and the processor.

majewsky:
The conformance-tests thing is always red. The problem is that there was a new version of the OCI Distribution Spec a while ago, and we  have not implemented the new features, so the respective tests just fail. It's known, and we keep putting it off because it's low prio.

PR #409, opened 2024-07-30
issue tokens with nbf 1 second in the past

majewsky:
We are seeing problems where Trivy rejects our tokens as "not valid yet", so there appears to be some clock drift going on. This might help alleviate the situation.


SuperSandro2000 (approved):
https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.5

PR #446, opened 2024-10-18
fix sublease token propagation during creation of managed replica accounts

majewsky:
A classic case of stringly-typed interface: Because the return type on the callback was only string, we did not notice that we returned the full serialized token when we should only have been returning the contents of the `secret` field. To avoid this error, the callback now returns a structured `keppel.SubleaseToken` value.

Also, I changed the callback to return just `error` instead of `*keppel.RegistryV2Error`. The latter is more consistent with what `CreateOrUpdateAccount` returns, but this change makes the callback implementations shorter, which overall helps readability.


SuperSandro2000 (approved):
Lgtm

PR #447, opened 2024-10-21
Backend deletion

SuperSandro2000:


docs/api-spec.md
```diff
@@ -142,8 +138,7 @@ The following fields may be returned:
...
 | `accounts[].gc_policies[].action` | string | One of: `delete` (to delete matching images) or `protect` (to not delete matching images, even if another policy with a lower priority would want to). |
-| `accounts[].in_maintenance` | bool | Whether this account is in maintenance mode. [See below](#maintenance-mode) for details. |
-| `accounts[].metadata` | object of strings | Free-form metadata maintained by the user. The contents of this field are not interpreted by Keppel, but may trigger special behavior in applications using this API. |
+| `accounts[].state` | string | The state of the account. Possible values are right now deleting [See below](#account-state) for details. |
```
majewsky, line 141: This reads like we're in the process of deleting possible values.
```suggestion
| `accounts[].state` | string | The state of the account. Only shown when there is a specific state to report. [See below](#account-state) for possible values and details. |
```

docs/api-spec.md
```diff
@@ -204,18 +199,16 @@ The following fields are shown on accounts configured with this strategy:
...
 
 - For primary accounts (i.e. accounts that are not replicas), no new blobs or manifests may be pushed. Only pulling and
   deleting are allowed.
 - For replica accounts, no new blobs or manifests will be replicated. Pulling is still allowed, but it becomes possible
   to delete blobs and manifests.
 
-Maintenance mode is a significant part of the account deletion workflow: Sending a DELETE request on an account is only
-allowed while the account is in maintenance mode, and the caller must have deleted all manifests from the account before
-attempting to DELETE it.
+Sending a DELETE request on an account marks it as `state = "deleting"` and the janitor starts to delete everything that belongs to the account including manifests and blobs.
```
majewsky, line 211: The janitor is an implementation detail that does not need to be exposed to users:

```suggestion
Sending a DELETE request on an account moves it into `state = "deleting"` and schedules the deletion of everything that belongs to the account, including manifests and blobs.
```

internal/api/keppel/accounts.go
```diff
@@ -158,18 +158,22 @@ func (a *API) handleDeleteAccount(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	resp, err := a.processor().DeleteAccount(r.Context(), *account, keppel.AuditContext{
+	if account.IsDeleting {
+		respondwith.JSON(w, http.StatusConflict, struct {
+			Error string `json:"error"`
+		}{Error: "account is already set to be deleted"})
+		return
```
majewsky, line 165: DELETE is usually an idempotent action in REST APIs, so we don't need to make this an error:
```suggestion
		w.WriteHeader(http.StatusNoContent)
		return
```

internal/api/keppel/accounts_test.go
```diff
@@ -300,40 +272,51 @@ func TestAccountsAPI(t *testing.T) {
...
 
 	// check editing of InMaintenance flag (this also tests editing of GC policies
 	// since we don't give any and thus clear the field)
-	for _, inMaintenance := range []bool{true, false} {
+	for _, isDeleting := range []bool{true, false} {
+		account := assert.JSONObject{
+			"auth_tenant_id": "tenant1",
+			"rbac_policies":  rbacPoliciesJSON,
+		}
+		accountExcept := assert.JSONObject{
```
majewsky, line 291:
```suggestion
		accountExpect := assert.JSONObject{
```

internal/api/keppel/accounts_test.go
```diff
@@ -300,40 +272,51 @@ func TestAccountsAPI(t *testing.T) {
...
+			"name":           "second",
+			"auth_tenant_id": "tenant1",
+			"rbac_policies":  rbacPoliciesJSON,
+		}
+		if isDeleting {
+			account = assert.JSONObject{
+				"auth_tenant_id": "tenant1",
+				"state":          "deleting",
```
majewsky, line 299: How is this test supposed to work? In any case, writing to `state` in the PUT endpoint should be rejected.

internal/api/keppel/accounts_test.go
```diff
@@ -1933,138 +1870,32 @@ func TestDeleteAccount(t *testing.T) {
...
-		},
-	}.Check(t, h)
-	_, err = s.DB.Exec(`UPDATE accounts SET in_maintenance = TRUE`)
+	_, err = s.DB.Exec(`UPDATE accounts SET is_deleting = FALSE`)
 	if err != nil {
 		t.Fatal(err.Error())
 	}
 
-	// phase 1: DELETE on account should complain about remaining manifests
+	// phase 1: DELETE on account should immeadetly mark it for deletion
```
majewsky, line 1878: Please double-check for typos:
```suggestion
	// phase 1: DELETE on account should immediately mark it for deletion
```

internal/keppel/account.go
```diff
@@ -30,10 +27,9 @@ type Account struct {
 	Name              models.AccountName    `json:"name"`
 	AuthTenantID      string                `json:"auth_tenant_id"`
 	GCPolicies        []GCPolicy            `json:"gc_policies,omitempty"`
-	InMaintenance     bool                  `json:"in_maintenance"`
-	Metadata          map[string]string     `json:"metadata"`
```
majewsky, line 34: When we discussed this, I didn't realize that these fields are not `omitempty`, so it's always shown on all accounts. Because of that, I would like to keep them around, always set to `false` and `nil` respectively, at least until we have removed usage of these fields from the Elektra UI.

internal/keppel/database.go
```diff
@@ -304,6 +304,28 @@ var sqlMigrations = map[string]string{
...
+	"043_add_accounts_is_deleting.down.sql": `
+    ALTER TABLE accounts
+      DROP COLUMN is_deleting,
+      DROP COLUMN next_deletion_attempt_at;
+
+    UPDATE accounts SET in_maintenance = TRUE WHERE is_deleting;
+
+    ALTER TABLE accounts
+      ADD COLUMN in_maintenance BOOLEAN NOT NULL DEFAULT FALSE,
+      ADD COLUMN metadata_json TEXT NOT NULL DEFAULT '';
```
majewsky, line 327: This order of steps will not work.

majewsky (commented):
I have read up until `internal/processor` (so have not looked at `internal/tasks` yet), but this ought to be enough for a first round.

Besides the notes below, in the Keppel API, all PUT/POST endpoints should return a client error (probably 409 Conflict) if `account.IsDeleting == true`. This can be done generically in `findAccountFromRequest` since `r.Method` is visible in that function.

internal/tasks/account_deletion.go
```diff
@@ -0,0 +1,213 @@
...
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	if len(nextManifests) > 0 {
+		// deletion was not completed yet -> check if we need to do something on our side
+		logg.Info("cleaning up account %q: need to delete %d manifests in this cycle", account.Name, len(nextManifests))
+		for _, rm := range nextManifests {
```
majewsky, line 128: This loop can be inlined with the previous one. Instead of collecting into `nextManifests`, each manifest can be deleted immediately.

internal/tasks/account_deletion.go
```diff
@@ -0,0 +1,213 @@
...
+			}
+			err = j.processor().DeleteManifest(ctx, account.Reduced(), *repo, parsedDigest, actx)
+			if err != nil {
+				return fmt.Errorf("while deleting manifest %q in repository %q: %w",
+					rm.Digest, rm.RepositoryName, err)
+			}
+		}
+	}
+
+	// delete all repos (and therefore, all blob mounts), so that blob sweeping can immediately take place
```
majewsky, line 147: After deleting manifests found by `deleteAccountFindManifestsQuery`, you move directly into deleting repos. But at this point, it's not a given that you actually deleted all manifests. The query only deletes manifests not referenced by a parent manifest. To actually get rid of all manifests, you need to check `SELECT COUNT(*)` of the remaining manifests in the account, and restart from the top if there's still stuff left.

This does not come up in the tests because we don't have parent-child relationships there (like in multi-arch images). To exercise this, the managed account deletion test could be expanded by also pushing a list manifest that references the existing image manifest.

internal/tasks/account_deletion.go
```diff
@@ -0,0 +1,213 @@
...
+		// TODO: how to prevent resetting time stamp if already set?
+		_, err := j.db.Exec(deleteAccountMarkAllBlobsForDeletionQuery, account.Name, j.timeNow())
+		if err != nil {
+			return err
+		}
+		_, err = j.db.Exec(deleteAccountScheduleBlobSweepQuery, account.Name, j.timeNow())
+		if err != nil {
+			return err
+		}
+		logg.Info("cleaning up managed account %q: waiting for %d blobs to be deleted", account.Name, blobCount)
```
majewsky, line 169: When deferring into a different job, you should update this job's clock field (`next_deletion_attempt_at`) by a little bit (e.g. 1 minute) to avoid this job getting hammered while it cannot make any progress.

internal/tasks/account_deletion.go
```diff
@@ -0,0 +1,213 @@
...
+	if userInfo := actx.UserIdentity.UserInfo(); userInfo != nil {
+		j.auditor.Record(audittools.EventParameters{
+			Time:       j.timeNow(),
+			Request:    actx.Request,
+			User:       userInfo,
+			ReasonCode: http.StatusOK,
+			Action:     cadf.DeleteAction,
+			Target:     processor.AuditAccount{Account: account},
+		})
+	}
```
majewsky, line 209: The user's DELETE already created an audit event. Since we're acting on behalf of the user here, a separate audit event is not necessary.

internal/tasks/account_deletion.go
```diff
@@ -0,0 +1,213 @@
...
+	deleteAccountScheduleBlobSweepQuery       = `UPDATE accounts SET next_blob_sweep_at = $2 WHERE name = $1`
+	deleteAccountMarkAllBlobsForDeletionQuery = `UPDATE blobs SET can_be_deleted_at = $2 WHERE account_name = $1`
+)
+
+type deleteAccountRemainingManifest struct {
+	RepositoryName string
+	Digest         string
+}
+
+func (j *Janitor) DeleteAccount(ctx context.Context, account models.Account, actx keppel.AuditContext) error {
```
majewsky, line 110: This can be a private method. In fact, I would just inline it with `deleteMarkedAccount` since that's the only callsite.

internal/tasks/account_management.go
```diff
@@ -97,16 +95,25 @@ func (j *Janitor) enforceManagedAccount(ctx context.Context, accountName models.
...
+		var accountModel *models.Account
+		accountModel, err = keppel.FindAccount(j.db, accountName)
+		if err != nil {
+			return err
+		}
+		if errors.Is(err, sql.ErrNoRows) {
+			nextCheckDuration = 0 // assume the account got already deleted
+		} else {
+			actx := keppel.AuditContext{
+				UserIdentity: janitorUserIdentity{TaskName: "account-sync"},
```
majewsky, line 107: I guess I didn't catch this when we introduced managed accounts, but `account-sync` is a weird name for this task.
```suggestion
				UserIdentity: janitorUserIdentity{TaskName: "managed-account-enforcement"},
```

internal/tasks/account_management.go
```diff
@@ -97,16 +95,25 @@ func (j *Janitor) enforceManagedAccount(ctx context.Context, accountName models.
...
+				Request:      janitorDummyRequest,
+			}
+			err = j.processor().MarkAccountForDeletion(*accountModel, actx)
+			if err == nil {
+				nextCheckDuration = 0 // account will be deleted -> next check not necessary
```
majewsky, line 112: 0 will not do what we want because that leaves the `next_enforcement_at` timestamp unchanged, causing this job to come back to this account immediately.
```suggestion
				nextCheckDuration = 1 * time.Hour // account will be deleted -> defer next check until probably after it was deleted
```

majewsky (changes requested):

internal/tasks/account_deletion.go
SuperSandro2000: nvm I found it

PR #450, opened 2024-10-29
various memory allocation optimizations

majewsky:
According to the debug/pprof/allocs profile that I pulled, this ought to reduce allocated memory in keppel-api by about 10%.


SuperSandro2000:
The conformance test is doing weird things https://github.com/sapcc/keppel/actions/runs/11571490340/job/32209485683?pr=450#step:5:13

SuperSandro2000 (approved):

PR #452, opened 2024-11-05
drivers/openstack: use more compact serialization for Keystone tokens

majewsky:
Since the new format should have all redundancies eliminated from it, I figure that the roundtrip through gzip and base64 is not going to save much space (or could actually waste bytes if gzip cannot save more than the 33% that base64 adds). In any case, it's more computationally expensive, so I'm getting rid of it.


SuperSandro2000 (approved):
We don't have benchmarks, do we?

majewsky:
> We don't have benchmarks, do we?

Not in the automated sense, no. Running a benchmark on this (in the https://pkg.go.dev/testing sense) would require instantiating the Keystone AuthPlugin, which would require having an OpenStack in scope for the tests. I will have a look at the CPU/RAM usage metrics after this is merged.

PR #482, opened 2025-01-17
Replace github.com/docker/distribution with github.com/containers/ima…

SuperSandro2000:
…ge/v5


go.mod
```diff
@@ -5,6 +5,7 @@ go 1.23
 require (
 	github.com/alicebob/miniredis/v2 v2.34.0
 	github.com/aquasecurity/trivy v0.58.2
+	github.com/containers/image/v5 v5.33.1
 	github.com/databus23/goslo.policy v0.0.0-20210929125152-81bf2876dbdb
 	github.com/dlmiddlecote/sqlstats v1.0.2
 	github.com/docker/distribution v2.8.3+incompatible
```
SuperSandro2000, line 11: That line should go away

internal/client/validate.go
```diff
@@ -99,18 +99,18 @@ func (c *RepoClient) doValidateManifest(ctx context.Context, reference models.Ma
...
 	}
-	manifest, manifestDesc, err := keppel.ParseManifest(manifestMediaType, manifestBytes)
+	manifest, err := keppel.ParseManifest(manifestMediaType, manifestBytes)
 	if err != nil {
 		return err
 	}
 
 	// the manifest itself looks good...
-	session.Logger.LogManifest(models.ManifestReference{Digest: manifestDesc.Digest}, level, nil, false)
+	session.Logger.LogManifest(models.ManifestReference{Digest: reference.Digest}, level, nil, false)
```
majewsky, line 108: `reference.Digest` will be empty if only a tag name is provided.

internal/keppel/manifest.go
```diff
@@ -24,158 +24,235 @@ import (
...
-	v1 "github.com/opencontainers/image-spec/specs-go/v1"
-
-	// distribution.UnmarshalManifest() relies on the following packages
-	// registering their manifest schemas.
-	"github.com/docker/distribution/manifest/manifestlist"
-	"github.com/docker/distribution/manifest/ocischema"
-	"github.com/docker/distribution/manifest/schema2"
+	"github.com/containers/image/v5/manifest"
+	"github.com/containers/image/v5/types"
+	imageSpecs "github.com/opencontainers/image-spec/specs-go/v1"
```
majewsky, line 29: Camel case is not commonly used for import names:
```suggestion
	imagespecs "github.com/opencontainers/image-spec/specs-go/v1" 
```

internal/keppel/manifest.go
```diff
@@ -24,158 +24,235 @@ import (
...
-	}
-	switch m := m.(type) {
-	case *schema2.DeserializedManifest:
-		return v2ManifestAdapter{m}, desc, nil
-	case *ocischema.DeserializedManifest:
-		return ociManifestAdapter{m}, desc, nil
-	case *manifestlist.DeserializedManifestList:
-		return listManifestAdapter{m}, desc, nil
+func ParseManifest(mediaType string, contents []byte) (ParsedManifest, error) {
+	// update ManifestMediaTypes if any new are added
```
majewsky, line 63: This looks like a description of the code, instead of an admonishment, so it will likely be missed at the important moments.
```suggestion
	// WARNING: Please update ManifestMediaTypes if any new are added.
```

internal/keppel/manifest.go
```diff
@@ -24,158 +24,235 @@ import (
...
 		// compatible with the reference implementation of Docker Hub, we serve this case by recursing
 		// into the image list and returning the linux/amd64 manifest to the client.
 		//
 		// This case is relevant for the support of tagged multi-arch images in `docker pull`.
-		if a.m.Versioned.MediaType == manifestlist.MediaTypeManifestList && m.MediaType == schema2.MediaTypeManifest {
+		if a.m.MediaType == manifest.DockerV2ListMediaType && m.MediaType == manifest.DockerV2Schema2MediaType {
 			if m.Platform.OS == "linux" && m.Platform.Architecture == "amd64" {
 				result = append(result, m)
 			}
 		}
```
majewsky, line 220: This `if` branch is unreachable on this type because this is a function on `ociIndexAdapter`.

internal/processor/manifests.go
```diff
@@ -174,7 +174,7 @@ func (p *Processor) ValidateExistingManifest(ctx context.Context, account models
 		return err
 	}
 
-	return p.validateAndStoreManifestCommon(ctx, account, repo, manifest, manifestBytes,
+	return p.validateAndStoreManifestCommon(ctx, account, repo, manifest, manifest.Digest, manifestBytes,
```
majewsky, line 177: By passing `manifest.Digest` here, you are neutering the digest check in `validateAndStoreManifestCommon` because it checks `manifest.Digest == manifest.Digest`, which can never fail.

You need to actually compute the digest from `manifestBytes` here. Otherwise, the storage corruption check portion of ValidateManifestsJob is disabled.

I actually recommend removing the `digest` argument entirely from `validateAndStoreManifestCommon`, and having it compute the digest from `manifestBytes` (like `distribution.UnmarshalManifest()` used to do).

majewsky (changes requested):
First batch:

internal/keppel/manifest.go
```diff
@@ -24,158 +24,235 @@ import (
...
+	"github.com/containers/image/v5/manifest"
+	"github.com/containers/image/v5/types"
+	imageSpecs "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
-//NOTE: We don't enable github.com/docker/distribution/manifest/schema1
-// anymore since it's legacy anyway and the implementation is a lot simpler
-// when we don't have to rewrite manifests between schema1 and schema2.
-
+// TODO: use manifest.Manifest instead!?!?
```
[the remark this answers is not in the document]
SuperSandro2000: I don't think we can do that since we need a few custom functions.

internal/processor/manifests.go
SuperSandro2000: > I actually recommend removing the `digest` argument entirely from `validateAndStoreManifestCommon`, and having it compute the digest from `manifestBytes` (like `distribution.UnmarshalManifest()` used to do).

Then we calculate it in ValidateAndStoreManifest twice.

internal/client/validate.go
```diff
@@ -99,18 +99,20 @@ func (c *RepoClient) doValidateManifest(ctx context.Context, reference models.Ma
 	if err != nil {
 		return err
 	}
-	manifest, manifestDesc, err := keppel.ParseManifest(manifestMediaType, manifestBytes)
+	manifest, err := keppel.ParseManifest(manifestMediaType, manifestBytes)
 	if err != nil {
 		return err
 	}
 
+	digest := digest.FromBytes(manifestBytes)
```
majewsky, line 107: If `reference` contains a digest, it should be checked if this matches.

internal/processor/manifests.go
```diff
@@ -184,24 +184,21 @@ type validateAndStoreManifestOpts struct {
 	ActionBeforeCommit func(*gorp.Transaction) error
 }
 
-func (p *Processor) validateAndStoreManifestCommon(ctx context.Context, account models.ReducedAccount, repo models.Repository, manifest *models.Manifest, manifestBytes []byte, opts validateAndStoreManifestOpts) error {
+func (p *Processor) validateAndStoreManifestCommon(ctx context.Context, account models.ReducedAccount, repo models.Repository, manifest *models.Manifest, digest digest.Digest, manifestBytes []byte, opts validateAndStoreManifestOpts) error {
```
majewsky, line 187: It still doesn't feel right to me that this interface does not encode that `digest` is (and must be) computed from `manifestBytes`. I propose adding a new type somewhere like

```
type BytesWithDigest struct {
  bytes []byte
  ownDigest digest.Digest 
}

func NewBytesWithDigest(buf []byte) BytesWithDigest {}
func (b BytesWithDigest) Bytes() []byte {}
func (b BytesWithDigest) Digest() digest.Digest {}
func (b BytesWithDigest) Len() int {}
```
to make the relationship clearer.

majewsky (changes requested):
I actually got through everything this time. \o/

majewsky (approved):

SuperSandro2000:
nope

majewsky (approved):

PR #483, opened 2025-01-23
Drop deprecated in_maintenance

SuperSandro2000:


internal/api/keppel/accounts_test.go
```diff
@@ -423,30 +413,25 @@ func TestPutAccountInMaintenanceFlag(t *testing.T) {
...
+			"account": assert.JSONObject{
+				"name":           "first",
+				"auth_tenant_id": "tenant1",
+				"metadata":       nil,
+				"rbac_policies":  []assert.JSONObject{},
 			},
-		}.Check(t, h)
-	}
+		},
+	}.Check(t, h)
```
majewsky, line 434: This test is called TestPutAccountInMaintenanceFlag, so it can go entirely.

internal/api/keppel/accounts_test.go
```diff
@@ -1161,21 +1143,6 @@ func TestPutAccountErrorCases(t *testing.T) {
 		ExpectBody:   assert.StringData("no permission for keppel_account:unknown:change\n"),
 	}.Check(t, h)
 
-	// TODO: reenable once we completely remove support for the accounts.in_maintenance flag
```
majewsky, line 1164: As the comment says, this test is relevant and should be reenabled instead of being deleted.

majewsky (changes requested):

majewsky (approved):

PR #500, opened 2025-02-21
Implement referrer api

SuperSandro2000:


majewsky (commented):
I have looked at everything except for the `internal/api` part, and that stuff looks good. Can you please move the DB model changes and the stuff that fills these fields into a separate PR? i.e. everything not in `internal/api`

majewsky (changes requested):
I pushed some commits to fix nitpicks; otherwise the implementation looks good. Please also add test coverage for the new endpoint, then we're good to go here.

PR #517, opened 2025-03-28
Switch to typos

SuperSandro2000:
Misspell doesn't find that much typos. [Typos](https://github.com/crate-ci/typos) is an alternative that finds more and generally has a low false positive rate.

I found two lala things in here: DELETEs and the rsa keys are not fully detected correctly but I think that doesn't outweigh the advantages.


majewsky (approved):
I think I'm fine with this in general. Do we want to have the `.typos.toml` autogenerated? The contents look highly specific to this repo. Maybe we generate one on first setup with a link to the documentation and a default set of excludes (`go.mod` and `vendor`) and then leave it to the repo owner to maintain it from that point?

SuperSandro2000:
Without templating support? We can do that

majewsky (approved):

PR #537, opened 2025-05-16
Misc header reflector and rate limiting stuff

SuperSandro2000:


wagnerd3 (approved):
Semantically, no issues from my side.

internal/keppel/ratelimit_driver.go
```diff
@@ -81,37 +81,25 @@ type RateLimitEngine struct {
 
 // RateLimitAllows checks whether the given action on the given account is allowed by
 // the account's rate limit.
-func (e RateLimitEngine) RateLimitAllows(ctx context.Context, remoteAddr string, account models.ReducedAccount, action RateLimitedAction, amount uint64) (bool, *redis_rate.Result, error) {
+func (e RateLimitEngine) RateLimitAllows(ctx context.Context, remoteAddr string, account models.ReducedAccount, action RateLimitedAction, amount uint64) (allowed bool, retryAfter time.Duration, err error) {
```
majewsky, line 84: I disagree with this being a net-positive refactor. All this does is save three lines of initializing fields that are not currently used, but might be used in the future (e.g. if the usecase arises to report `result.Remaining` in headers later). In general, I prefer more structured returns over less structured ones, so giving `*redis_rate.Result` instead of `time.Duration` seems better to me, and it does not appear to be constraining us to continue doing so.

majewsky (changes requested):

wagnerd3 (approved):
Whoops, suddenly it's a copyright PR :D

majewsky:
I put the license header changes on `master` directly. Please rebase.

SuperSandro2000:
The tests actually found a regression that allowed in result was not set.

majewsky (approved):

PR #538, opened 2025-05-19
Tag policy

SuperSandro2000:


internal/api/utils.go
```diff
@@ -40,3 +41,7 @@ func CheckRateLimit(r *http.Request, rle *keppel.RateLimitEngine, account models
 
 	return nil
 }
+
+var GetTagPolicyByAccountName = sqlext.SimplifyWhitespace(`
```
majewsky, line 45:
```suggestion
var GetTagPolicyByAccountNameQuery = sqlext.SimplifyWhitespace(`
```

internal/api/keppel/manifests_test.go
```diff
@@ -43,18 +43,8 @@ func TestManifestsAPI(t *testing.T) {
...
-			SecurityScanPoliciesJSON: "[]",
-		})
-		mustInsert(t, s.DB, &models.Account{
-			Name:                     "test2",
-			AuthTenantID:             "tenant2",
-			GCPoliciesJSON:           "[]",
-			SecurityScanPoliciesJSON: "[]",
-		})
+		test.MustExec(t, s.DB, `INSERT INTO accounts (name, auth_tenant_id) VALUES ('test1', 'tenant1')`)
+		test.MustExec(t, s.DB, `INSERT INTO accounts (name, auth_tenant_id) VALUES ('test2', 'tenant2')`)
```
majewsky, line 47: We can also use `test.WithAccount()` for this, no? That appears to have the logic to default the `PoliciesJSON` fields to `[]`, too.

internal/api/keppel/repos_test.go
```diff
@@ -17,39 +17,13 @@ import (
...
-		SecurityScanPoliciesJSON: "[]",
-	})
-	mustInsert(t, s.DB, &models.Account{
-		Name:                     "test2",
-		AuthTenantID:             "tenant2",
-		GCPoliciesJSON:           "[]",
-		SecurityScanPoliciesJSON: "[]",
-	})
+	test.MustExec(t, s.DB, `INSERT INTO accounts (name, auth_tenant_id) VALUES ('test1', 'tenant1')`)
+	test.MustExec(t, s.DB, `INSERT INTO accounts (name, auth_tenant_id) VALUES ('test2', 'tenant2')`)
```
majewsky, line 26: Same here.

internal/api/registry/manifests.go
```diff
@@ -72,7 +72,16 @@ func (a *API) handleGetOrHeadManifest(w http.ResponseWriter, r *http.Request) {
...
 
-			dbManifest, manifestBytes, err = a.processor().ReplicateManifest(r.Context(), *account, *repo, reference, keppel.AuditContext{
+			tagPoliciesStr, err := a.db.SelectStr(api.GetTagPolicyByAccountName, account.Name)
+			if respondWithError(w, r, err) {
+				return
+			}
+			tagPolicies, err := keppel.ParseTagPolicies(tagPoliciesStr)
+			if respondWithError(w, r, err) {
+				return
+			}
```
majewsky, line 82: Since we have the query in `internal/api/utils.go` already, can we put all of this (take account name, pull JSON from DB and parse) as a helper function there?

internal/tasks/image_gc.go
```diff
@@ -227,25 +231,25 @@ func (j *Janitor) evaluatePolicy(ctx context.Context, proc *processor.Processor,
...
-		pCopied := policy
+		pCopied := gcPolicies
 		// execute policy action
-		switch policy.Action {
+		switch gcPolicies.Action {
 		case "protect":
 			m.GCStatus.ProtectedByPolicy = &pCopied
 		case "delete":
-			err := proc.DeleteManifest(ctx, account, repo, m.Manifest.Digest, keppel.AuditContext{
+			err := proc.DeleteManifest(ctx, account, repo, m.Manifest.Digest, tagPolicies, keppel.AuditContext{
```
majewsky, line 252: This deletion should not be attempted if the tag policies forbid it. Otherwise image GC will get stuck on a useless error.

Instead, the manifest's GCStatus should gain a new field `ProtectedByTagPolicy` that gets filled in this case (which also means API visibility for this mechanism).

To avoid duplicating the tag policy evaluation between here and Processor.DeleteManifest(), I recommend returning a defined error type from DeleteManifest() if a tag policy blocks deletion:

```
type DeleteManifestBlockedByTagPolicyError struct {
  Policy keppel.TagPolicy
}
```

and checking for this here via `errext.As`.

majewsky (changes requested):
Thanks for the effort! A lot of my notes are nitpicks, and also there is one bigger remark near the end about the API surface.

internal/tasks/image_gc.go
```diff
@@ -227,40 +232,43 @@ func (j *Janitor) evaluatePolicy(ctx context.Context, proc *processor.Processor,
...
 				UserIdentity: janitorUserIdentity{
 					TaskName: "policy-driven-gc",
 					GCPolicy: &pCopied,
 				},
 				Request: janitorDummyRequest,
 			})
+			if _, ok := errext.As[processor.DeleteManifestBlockedByTagPolicyError](err); ok { //nolint:errcheck // intended to be skipped
+				continue
+			}
```
majewsky, line 262: The error should not be swallowed. As described in my previous review notes, type keppel.GCStatus should be extended to report the relevant tag policy.

internal/tasks/image_gc.go
```diff
@@ -207,7 +212,7 @@ outer:
 	return j.persistGCStatus(manifests, repo.ID)
 }
 
-func (j *Janitor) evaluatePolicy(ctx context.Context, proc *processor.Processor, manifests []*manifestData, account models.ReducedAccount, repo models.Repository, policy keppel.GCPolicy) error {
+func (j *Janitor) evaluatePolicy(ctx context.Context, proc *processor.Processor, manifests []*manifestData, account models.ReducedAccount, repo models.Repository, gcPolicies keppel.GCPolicy, tagPolicies []keppel.TagPolicy) error {
```
majewsky, line 215: It's only one GC policy here:
```suggestion
func (j *Janitor) evaluatePolicy(ctx context.Context, proc *processor.Processor, manifests []*manifestData, account models.ReducedAccount, repo models.Repository, gcPolicy keppel.GCPolicy, tagPolicies []keppel.TagPolicy) error {
```

majewsky (changes requested):
Regarding the problem that `keppel.RegistryV2Error` cannot hold a structured error, I pushed a commit that fixes that. Please take a look. Also:

internal/keppel/gc_policy.go
```diff
@@ -245,6 +214,8 @@ type GCStatus struct {
 	// If the image is not protected, contains all policies with action "delete"
 	// that could delete this image in the future.
 	RelevantPolicies []GCPolicy `json:"relevant_policies,omitempty"`
+	// If the image is protected, contains all tag policies that protect this image.
+	ProtectedByTagPolicies []TagPolicy `json:"protected_by_tag_policies,omitempty"`
```
majewsky, line 218: This should only show the one tag policy that actually blocked the deletion:

```suggestion
	ProtectedByTagPolicy Option[TagPolicy] `json:"protected_by_tag_policy,omitempty"`
```

majewsky (changes requested):

majewsky (approved):
Thanks a bunch!

PR #546, opened 2025-05-27
use Option[] instead of pointers in several places

majewsky:
Despite the branch name, this is far from all instances of pointer usage where an Option should be used. But it's a start. I decided to abort here because I got pulled into the deep end, and it's already way past the end of the workday.

### DIFF
--- a/internal/keppel/utils.go
+++ b/internal/keppel/utils.go
@@ -8,6 +8,8 @@ import (
 	"net/url"
 	"strings"
 	"time"
+
+	. "github.com/majewsky/gg/option"
 )
 
 // OriginalRequestURL returns the URL that the original requester used when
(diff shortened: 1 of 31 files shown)


internal/keppel/utils.go
```diff
@@ -49,38 +51,44 @@ func AppendQuery(urlStr string, query url.Values) string {
 }
 
 // MaybeTimeToUnix casts a time.Time instance into its UNIX timestamp while preserving nil-ness.
```
SuperSandro2000, line 53: Should we update this comment, too?

internal/tasks/manifests.go
```diff
@@ -621,8 +622,8 @@ func (j *Janitor) doSecurityCheck(ctx context.Context, securityInfo *models.Triv
 
 	// clear timing information (this will be filled down below once we actually talk to Trivy;
 	// if any preflight check fails, the fields stay at nil)
-	securityInfo.CheckedAt = nil
```
SuperSandro2000, line 624: Above comment should probably say none

SuperSandro2000 (approved):

internal/keppel/utils.go
```diff
@@ -48,39 +50,45 @@ func AppendQuery(urlStr string, query url.Values) string {
 	return urlStr + "?" + query.Encode()
 }
 
-// MaybeTimeToUnix casts a time.Time instance into its UNIX timestamp while preserving nil-ness.
-func MaybeTimeToUnix(t *time.Time) *int64 {
-	if t == nil {
-		return nil
+// MaybeTimeToUnix casts a time.Time instance into its UNIX timestamp unless it is None.
+func MaybeTimeToUnix(t Option[time.Time]) Option[int64] {
```
wagnerd3, line 54: Newbie perspective: in a function name the `maybe` sounds like "the function might convert your time to Unix, but sometimes it might not". I tried to read up on it and found the sometimes an `Opt`/ `Option` suffix or Prefix is used instead.
E.g. `TimeOptToUnix`, `OptTimeToUnix`, `TimeOptionToUnix`, `OptionTimeToUnix`

I do get the point, that the way you chose is consistent with the `maybe...` variables. If it's just me, disregard this comment :-)

wagnerd3 (approved):

PR #565, opened 2025-07-17
Skip trivy security scan for accounts that are in deletion

SuperSandro2000:


internal/tasks/manifests.go
```diff
@@ -698,6 +702,16 @@ func (j *Janitor) doSecurityCheck(ctx context.Context, securityInfo *models.Triv
...
 		}
 		securityStatuses = append(securityStatuses, status)
+
+		// abort if the account is in deletion mode in the meantime
+		accountIsInDeleting, err := j.db.SelectBool(`SELECT is_deleting FROM accounts WHERE name = $1`, account.Name)
+		if err != nil {
+			return fmt.Errorf("could not check if account %s is in deletion mode: %w", account.Name, err)
+		}
+		if accountIsInDeleting {
+			return fmt.Errorf("account %s is in deletion mode, skik storing Trivy report", account.Name)
```
majewsky, line 712:
```suggestion
			return fmt.Errorf("account %s is in deletion mode, skip storing Trivy report", account.Name)
```

PR #573, opened 2025-07-31
replace validation.required_labels with CEL expressions

Varsius:


internal/keppel/validation.go
```diff
@@ -6,37 +6,109 @@ package keppel
...
+	// accept if and only if the presented CEL expression is the same as what we would generate from the presented list of required labels
+	if v.RuleForManifest != "" && v.RequiredLabels != nil {
+		generatedRule := generateRuleForManifest(v.RequiredLabels)
+		if generatedRule != v.RuleForManifest {
+			err := fmt.Errorf(`required labels %q do not match rule for manifest %q`, v.RequiredLabels, v.RuleForManifest)
+			return AsRegistryV2Error(err).WithStatus(http.StatusUnprocessableEntity)
+		}
+	}
+	if v.RuleForManifest != "" {
+		_, ast, celErr := BuildManifestValidationAST(v.RuleForManifest)
```
SuperSandro2000, line 56: Should we cache that in some kind or LRU?

internal/keppel/validation.go
```diff
@@ -6,37 +6,109 @@ package keppel
...
 
-	return &ValidationPolicy{
-		RequiredLabels: account.SplitRequiredLabels(),
+	policy := ValidationPolicy{
+		RuleForManifest: account.RuleForManifest,
+	}
+
+	// for backwards compatibility, show required_labels field if the CEL expression is identical
+	// to the one that we would generate from a required_labels attribute
+	re := regexp.MustCompile(`^has\(labels\.([a-zA-Z_][a-zA-Z0-9_]*)\)(\s*&&\s*has\(labels\.([a-zA-Z_][a-zA-Z0-9_]*)\))*$`)
```
SuperSandro2000, line 36: We should move this out of the function onto a var to only compile it once.

internal/api/registry/manifests_test.go
```diff
@@ -600,14 +600,14 @@ func TestManifestRequiredLabels(t *testing.T) {
 			ExpectHeader: test.VersionHeader,
 			ExpectBody: test.ErrorCodeWithMessage{
 				Code:    keppel.ErrManifestInvalid,
-				Message: "missing required labels: somethingelse, andalsothis",
+				Message: "rule \"has(labels.randomlabelthatdoesnotexist)\" is not satisfied",
```
SuperSandro2000, line 603: Can we somehow extract why the rule did not apply? Especially when combining boolean conditions it could be a bit confusing why exactly things failed.

internal/keppel/validation.go
majewsky: Yes, please use a https://pkg.go.dev/github.com/hashicorp/golang-lru/v2#Cache inside of BuildManifestValidationAST. It's enough to cache positive results.

internal/keppel/validation.go
```diff
@@ -6,37 +6,111 @@ package keppel
...
+	var propositions []string
+	for _, label := range requiredLabels {
+		propositions = append(propositions, fmt.Sprintf("has(labels.%s)", label))
+	}
+
+	return strings.Join(propositions, " && ")
+}
+
+func BuildManifestValidationAST(celExpression string) (*cel.Env, *cel.Ast, error) {
+	env, err := cel.NewEnv(cel.Variable("labels", cel.MapType(cel.StringType, cel.StringType)))
```
majewsky, line 107: If `*cel.Env` is thread-safe (i.e. it is safe for us to use the same instance across multiple goroutines), then this can be a global variable that's instantiated once (with the error handling done through `must.Return()`).

The documentation does not explicitly state thread-safety, but [the implementation](https://github.com/google/cel-go/blob/v0.26.0/cel/env.go#L131) holds mutexes, so it appears to be designed that way.

internal/processor/manifests.go
```diff
@@ -224,18 +224,26 @@ func (p *Processor) validateAndStoreManifestCommon(ctx context.Context, account
...
+		if validationRequired {
+			env, ast, err := keppel.BuildManifestValidationAST(account.RuleForManifest)
+			if err != nil {
+				return err
+			}
+			prg, err := env.Program(ast)
+			if err != nil {
+				return err
+			}
+			out, _, err := prg.Eval(map[string]interface{}{
```
majewsky, line 239: Nitpick: (did we not have a lint for this?)
```suggestion
			out, _, err := prg.Eval(map[string]any{
```

internal/processor/manifests.go
```diff
@@ -224,18 +224,26 @@ func (p *Processor) validateAndStoreManifestCommon(ctx context.Context, account
...
+				"labels": configInfo.Labels,
+			})
+			if err != nil {
+				return err
 			}
-			if len(missingLabels) > 0 {
-				msg := "missing required labels: " + strings.Join(missingLabels, ", ")
+			if out != types.True {
+				msg := fmt.Sprintf("rule %q is not satisfied", account.RuleForManifest)
```
majewsky, line 246: For rules that are very long, this might read better as:
```suggestion
				msg := fmt.Sprintf("rule is not satisfied: %s", account.RuleForManifest)
```

internal/drivers/basic/account_management_test.go
```diff
@@ -74,7 +74,8 @@ func TestConfigureAccount(t *testing.T) {
 			},
 		},
 		ValidationPolicy: &keppel.ValidationPolicy{
-			RequiredLabels: []string{"important-label", "some-label"},
```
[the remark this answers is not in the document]
majewsky: Yes, I find that to be problematic. Please use the `x['y']` syntax instead.

internal/api/registry/manifests_test.go
majewsky: I think it's fine to just say "this rule was not satisified", and then the caller will have to figure out which part they did wrong. This is kind of unavoidable when going with generic expressions. Even the stuff that Valentin describes, while potentially useful for the classic `has(labels.foo) && has(labels.bar)` usecase, will fall apart with more sophisticated expressions. So we would have to do a lot of work to give out a diagnostic that only helps with reasonably simple cases anyway, which I don't find to be a useful tradeoff.

majewsky (changes requested):
I read the implementation and existing comments, but not the tests yet. But this should be enough for the first round anyway:

internal/tasks/account_management_test.go
```diff
@@ -42,7 +42,7 @@ func TestAccountManagementBasic(t *testing.T) {
...
+		INSERT INTO accounts (name, auth_tenant_id, external_peer_url, gc_policies_json, security_scan_policies_json, rbac_policies_json, is_managed, next_enforcement_at, rule_for_manifest) VALUES ('abcde', '12345', 'registry-tertiary.example.org', '[{"match_repository":".*/database","except_repository":"archive/.*","time_constraint":{"on":"pushed_at","newer_than":{"value":6,"unit":"h"}},"action":"protect"},{"match_repository":".*","only_untagged":true,"action":"delete"}]', '[{"match_repository":".*","match_vulnerability_id":".*","except_fix_released":true,"action":{"assessment":"risk accepted: vulnerabilities without an available fix are not actionable","ignore":true}}]', '[{"match_repository":"library/.*","permissions":["anonymous_pull"]},{"match_repository":"library/alpine","match_username":".*@tenant2","permissions":["pull","push"]}]', TRUE, %d, ''important-label' in labels && 'some-label' in labels');
```
majewsky, line 45: The quoting here is incorrect. I'll follow up on that with a fix in go-bits.

**Edit:** sapcc/go-bits#245

majewsky (changes requested):
Did not get very far reading this today; I was mostly dealing with Limes PRs today. I found a little nitpick though:

SuperSandro2000 (approved):

PR #587, opened 2025-09-01
bump all dependencies to their latest versions

majewsky:
This is the same as <https://github.com/sapcc/keppel/pull/582>, except for the changes that we want to ignore because go-makefile-maker is supposed to make them.

### DIFF
--- a/.github/workflows/oci-distribution-conformance.yml
+++ b/.github/workflows/oci-distribution-conformance.yml
@@ -27,7 +27,7 @@ jobs:
         uses: actions/setup-go@v5
         with:
           check-latest: true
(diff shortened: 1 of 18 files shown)


.github/workflows/oci-distribution-conformance.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         uses: actions/setup-go@v5
         with:
           check-latest: true
-          go-version: "1.24"
+          go-version: "1.25"
```
SuperSandro2000, line 30:
```suggestion
          go-version: "1.24"
```

PR #595, opened 2025-09-24
Add test case for required_labels on OCI indexes

SuperSandro2000:


internal/processor/manifests.go
```diff
@@ -410,6 +410,14 @@ type manifestConfigInfo struct {
 
 // Returns the list of missing labels, or nil if everything is ok.
 func parseManifestConfig(ctx context.Context, tx *gorp.Transaction, sd keppel.StorageDriver, account models.ReducedAccount, manifest keppel.ParsedManifest) (result manifestConfigInfo, err error) {
+	// image manifests produced by docker buildx are weird and have their labels **not** in the config like normally
+	// but in the manifest's annotations.
+	// In addition to that their config is uploaded late, so collecting history also fails...
+	if manifest.GetArtifactType() == imagespecs.MediaTypeImageManifest {
+		result.Labels = manifest.GetAnnotations()
+		return result, nil
```
majewsky, line 418: This looks like it breaks OCI images that have the labels in the config blob in the correct way. We should not return here, but rather merge both sets of labels together.

majewsky (changes requested):

PR #606, opened 2025-10-17
Remove allocs from io.ReadAll (hard cases)

SuperSandro2000:
Lets not merge this until Stefan gave a review.

### DIFF
--- a/internal/api/keppel/repos_test.go
+++ b/internal/api/keppel/repos_test.go
@@ -79,7 +79,7 @@ func TestReposAPI(t *testing.T) {
 			PushedAt:         manifestPushedAt,
 			NextValidationAt: manifestPushedAt.Add(models.ManifestValidationInterval),
 		}))
-		must.SucceedT(t, s.SD.WriteManifest(s.Ctx, models.ReducedAccount{Name: "test1"}, "repo1-3", dummyDigest, []byte("data")))
+		must.SucceedT(t, s.SD.WriteManifest(s.Ctx, models.ReducedAccount{Name: "test1"}, "repo1-3", dummyDigest, strings.NewReader("data")))
 		must.SucceedT(t, s.DB.Insert(&models.TrivySecurityInfo{
 			RepositoryID:        filledRepo.ID,
 			Digest:              dummyDigest,
--- a/internal/tasks/manifests.go
+++ b/internal/tasks/manifests.go
@@ -443,7 +443,12 @@ func (j *Janitor) collectManifestLayerBlobs(ctx context.Context, account models.
 	}
 
 	// we only care about blobs that are image layers; the manifest tells us which blobs are layers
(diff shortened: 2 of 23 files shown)


internal/keppel/security_scan_policy_test.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/sapcc/go-bits/must"
+)
+
+func BenchmarkEnrichReport(b *testing.B) {
+	for b.Loop() {
+		policies := SecurityScanPolicySet{}
+		content := must.Return(os.ReadFile("../tasks/fixtures/trivy/report-vulnerable-with-fixes.json"))
+		report := trivy.ReportPayload{
+			Format:   "json",
+			Contents: io.NopCloser(bytes.NewReader(content)),
```
majewsky: I feel that it would be more faithful to the benchmark to use `os.Open()` directly here:
```suggestion
		report := trivy.ReportPayload{
			Format:   "json",
			Contents: must.Return(os.Open("../tasks/fixtures/trivy/report-vulnerable-with-fixes.json")),
```
And then `must.Succeed(file.Close())` at the end of the loop. (This should not use `defer` because then the benchmark may be impacted by open file handles piling up across loop iterations.)

internal/tasks/manifests.go
```diff
@@ -443,7 +443,11 @@ func (j *Janitor) collectManifestLayerBlobs(ctx context.Context, account models.
 	}
 
 	// we only care about blobs that are image layers; the manifest tells us which blobs are layers
-	manifestBytes, err := j.sd.ReadManifest(ctx, account, repo.Name, manifest.Digest)
+	manifestReader, err := j.sd.ReadManifest(ctx, account, repo.Name, manifest.Digest)
```
majewsky: Does not call `manifestReader.Close()`.

majewsky (changes requested):
I don't like that we're just ignoring most of the errors from `Close()`. I would like to have helper functions `ReadAllAndClose(io.ReadCloser) ([]byte, error)` and `CopyAllAndClose(io.Writer, io.ReadCloser) error` instead.

internal/api/keppel/repos_test.go
```diff
@@ -79,7 +80,7 @@ func TestReposAPI(t *testing.T) {
 			PushedAt:         manifestPushedAt,
 			NextValidationAt: manifestPushedAt.Add(models.ManifestValidationInterval),
 		}))
-		must.SucceedT(t, s.SD.WriteManifest(s.Ctx, models.ReducedAccount{Name: "test1"}, "repo1-3", dummyDigest, []byte("data")))
+		must.SucceedT(t, s.SD.WriteManifest(s.Ctx, models.ReducedAccount{Name: "test1"}, "repo1-3", dummyDigest, bytes.NewReader([]byte("data"))))
```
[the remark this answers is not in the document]
SuperSandro2000: Used search and replace to find some more cases.

PR #615, opened 2025-11-12
accept configuration for drivers via JSON payload in environment variables

majewsky:
For example, for the AuthDriver `trivial`:

```sh
# before
export KEPPEL_DRIVER_AUTH=trivial
export KEPPEL_PASSWORD=SuperSecret
export KEPPEL_USERNAME=johndoe
# after
export KEPPEL_DRIVER_AUTH='{"type":"trivial","params":{"username":"johndoe","password":"SuperSecret"}}'
```

My main motivation for this change is that I want to add a new StorageDriver that writes into multiple actual StorageDrivers simultaenously to allow for a migration between StorageDrivers. When migrating existing Swift deployments to use the Swift API emulation of Ceph, we will use this migration driver together with two instances of the `swift` StorageDriver.

Currently, the `swift` StorageDriver would accept configuration from environment variables, but then the two different `swift` StorageDriver instances would use the same environment variables and thus not be able to accept configuration. By switching to structured data for driver configuration, we can have the migration driver instance take two full sets of configuration for the two inner driver instances.

Other types of drivers also benefit from this, esp. the RateLimitDriver configuration, which is currently scattered in the Helm chart across a sea of environment variables (see `KEPPEL_BURST_*` and `KEPPEL_RATELIMIT_*` in [the Helm chart templating](https://github.com/sapcc/helm-charts/blob/6f9df964f972d6ce52ef0456b3a45e60110b5609/openstack/keppel/templates/_utils.tpl)).

I left a few pieces of driver configuration in env vars, either because the library in question wants to read from envvars (`gophercloudext.NewProviderClient`), or because the configuration includes secrets, which in k8s are easiest to supply via env variables.

- [ ] TODO: companion helm-charts PR


internal/drivers/basic/ratelimit.go
```diff
@@ -5,71 +5,58 @@ package basic
...
-		"m": redis_rate.PerMinute,
-		"h": redis_rate.PerHour,
-	}
-)
-
 func init() {
-	keppel.RateLimitDriverRegistry.Add(func() keppel.RateLimitDriver {
-		return RateLimitDriver{make(map[keppel.RateLimitedAction]redis_rate.Limit)}
-	})
+	keppel.RateLimitDriverRegistry.Add(func() keppel.RateLimitDriver { return RateLimitDriver{} })
```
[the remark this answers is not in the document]
majewsky: Related to the incorrect reasoning above.

internal/drivers/openstack/keystone.go
```diff
@@ -67,8 +71,7 @@ func (d *keystoneDriver) Init(ctx context.Context, rc *redis.Client) (err error)
 
 	// load oslo.policy
 	d.TokenValidator = &gopherpolicy.TokenValidator{IdentityV3: d.IdentityV3}
-	policyFilePath := osext.MustGetenv("KEPPEL_OSLO_POLICY_PATH")
```
SuperSandro2000: Do we somehow check if this is still set and then abort?

internal/keppel/config.go
```diff
@@ -164,3 +166,39 @@ func GetRedisOptions(prefix string) (*redis.Options, error) {
...
+		// configJSON was just a type, e.g. `{"type":"unittest"}`
+		cfg.Params = json.RawMessage("{}")
+	}
+	logg.Debug("initializing %s %q", driverType, configJSON)
+
+	driver, ok := registry.TryInstantiate(cfg.PluginTypeID).Unpack()
+	if !ok {
+		return zero, fmt.Errorf("no such %s: %q", driverType, cfg.PluginTypeID)
+	}
+	err = json.Unmarshal([]byte(cfg.Params), driver)
```
SuperSandro2000: How about setting disallowUnkownFields?

conformance-test/env.sh
```diff
@@ -9,11 +9,11 @@ export KEPPEL_DB_CONNECTION_OPTIONS=sslmode=disable
 export KEPPEL_DB_PASSWORD=
 export KEPPEL_DB_PORT=54321
 
+KEPPEL_FILESYSTEM_PATH=./conformance-test/storage
```
SuperSandro2000: Why not export?

internal/keppel/config.go
majewsky: done

conformance-test/env.sh
majewsky: Because I don't want to create the impression that Keppel reads this variable. This is only used to DRY the shell script.

SuperSandro2000 (approved):

PR #633, opened 2025-12-11
Test Harbor+ghcr.io error code special handling

SuperSandro2000:


internal/client/auth_challenge.go
```diff
@@ -111,20 +108,20 @@ func (c AuthChallenge) GetToken(ctx context.Context, userName, password string)
...
-		} else {
-			errMsg = data.Errors[0].Message
+		return "", keppel.AsRegistryV2Error(fmt.Errorf("auth token request to %q did return: %w", req.URL, err))
+	case len(data.Errors) == 1:
+		return "", data.Errors[0]
+	case len(data.Errors) > 1:
+		var errStr strings.Builder
+		for _, e := range data.Errors {
+			errStr.WriteString(e.Error() + "; ")
 		}
```
majewsky: This produces an extra trailing `"; "` at the end. Please use `errext.ErrorSet` and then `errs.Join("; ")`.

PR #639, opened 2026-01-15
fix more regressions in package client, add test coverage

majewsky:
- Returning `*keppel.RegistryV2Error` from `AuthChallenge.GetToken()` does not have any benefit, but causes a
problem because the calling function can now return the non-nil empty error `error((*keppel.RegistryV2Error)(nil))`.
- The API endpoint for monolithic blob uploads requires a slash at the end of the path. This trailing slash is required by the API specification, see `end-4b` in <https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#endpoints>.

### DIFF
--- a/internal/client/repo_client.go
+++ b/internal/client/repo_client.go
@@ -11,7 +11,6 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"path"
 	"strings"
 
 	"maps"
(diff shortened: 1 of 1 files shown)


internal/client/upload.go
```diff
@@ -23,7 +23,8 @@ func (c *RepoClient) UploadMonolithicBlob(ctx context.Context, contents []byte)
 		Path:   "blobs/uploads/",
 		Query:  url.Values{"digest": []string{d.String()}},
 		Headers: http.Header{
-			"Content-Type": {"application/octet-stream"},
+			"Content-Length": {strconv.FormatUint(uint64(len(contents)), 10)},
```
majewsky: This header must now be added manually because the testcase does not go through a real HTTP server (where it would normally be added automatically).

PR #695, opened 2026-04-17
Fix trivial storage driver not containing AuthTenantID

SuperSandro2000:


internal/api/registry/shared_test.go
```diff
@@ -33,7 +33,7 @@ var (
 	barRepoRef = models.Repository{AccountName: "test1", Name: "bar"}
 )
 
-// the auth tenant ID that all test accounts use
+// the auth tenant ID that many test accounts use
```
SuperSandro2000, line 36: @majewsky should I migrate all tests to that authtenantid?

SuperSandro2000:
The extra nil pointer checks will redundant with #696

internal/api/registry/shared_test.go
majewsky: Please do.

majewsky (approved):

PR #696, opened 2026-04-17
Remove pointer that FindAccount/FindReducedAccount return

SuperSandro2000:


internal/api/keppel/api.go
```diff
@@ -188,8 +187,7 @@ func (a *API) findRepositoryFromRequest(w http.ResponseWriter, r *http.Request,
 	if errors.Is(err, sql.ErrNoRows) {
 		http.Error(w, "repository not found", http.StatusNotFound)
 		return nil
-	}
-	if respondwith.ObfuscatedErrorText(w, err) {
+	} else if respondwith.ObfuscatedErrorText(w, err) {
```
wagnerd3: I think for early return this is an anti-pattern to continue in else branches.

internal/api/registry/api.go
```diff
@@ -256,6 +253,8 @@ func (a *API) checkAccountAccess(w http.ResponseWriter, r *http.Request, strateg
 		// issue tokens with scopes for nonexistent accounts)
 		keppel.ErrNameUnknown.With("account not found").WriteAsRegistryV2ResponseTo(w, r)
 		return nil, nil, nil, nil
+	} else if respondWithError(w, r, err) {
```
wagnerd3: Same here, no `else if` after early return?

wagnerd3 (dismissed):
I think logic is fine

internal/api/keppel/api.go
SuperSandro2000: We need to differentiate err = sql.ErrNoRows and err being something else.

majewsky (approved):

PR #698, opened 2026-04-17
Remove Pointer slice types, Avoid extra buffer WriteAsRegistryV2ResponseTo

SuperSandro2000:


internal/auth/token.go
```diff
@@ -135,7 +139,7 @@ func (a Authorization) IssueTokenWithExpires(cfg keppel.Configuration, expiresIn
 			IssuedAt:  jwt.NewNumericDate(now),
 		},
 		// access permissions granted to this token
-		Access:   a.ScopeSet.Flatten(),
+		Access:   access,
 		Embedded: embeddedUserIdentity{UserIdentity: a.UserIdentity},
```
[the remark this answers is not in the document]
SuperSandro2000: but it compiles 🤨

wagnerd3 (dismissed):
Should be fine? Though I cannot really judge how much this saves us...

SuperSandro2000:
The ScopeSet benchmark test now requires 1 alloc instead of 3.

The json marshall at the end is slightly better:

```
Before:

 ➜ go test ./internal/keppel/ -bench BenchmarkWriteAsRegistryV2ResponseTo
goos: linux
goarch: amd64
pkg: github.com/sapcc/keppel/internal/keppel
cpu: Intel(R) Xeon(R) Platinum 8260 CPU @ 2.40GHz
BenchmarkWriteAsRegistryV2ResponseTo-8            271560              4221 ns/op            1352 B/op         16 allocs/op
PASS
ok      github.com/sapcc/keppel/internal/keppel 1.163s

After:
 ➜ go test ./internal/keppel/ -bench BenchmarkWriteAsRegistryV2ResponseTo
goos: linux
goarch: amd64
pkg: github.com/sapcc/keppel/internal/keppel
cpu: Intel(R) Xeon(R) Platinum 8260 CPU @ 2.40GHz
BenchmarkWriteAsRegistryV2ResponseTo-8            261852              4599 ns/op            1225 B/op         16 allocs/op
PASS
ok      github.com/sapcc/keppel/internal/keppel 1.232s
```

I did not test the response type pointer but I expect it to be similar.

internal/keppel/errors.go
```diff
@@ -161,12 +161,12 @@ func (e *RegistryV2Error) WriteAsRegistryV2ResponseTo(w http.ResponseWriter, r *
...
 	}
 	if r.Method != http.MethodHead {
-		buf, _ := json.Marshal(struct {
+		//nolint:errcheck // we can't do much about errors at this point, and the client will see an incomplete response if we fail to write the error body, so we ignore them
+		_ = json.NewEncoder(w).Encode(struct {
 			Errors []*RegistryV2Error `json:"errors"`
 		}{
 			Errors: []*RegistryV2Error{e},
 		})
-		w.Write(append(buf, '\n'))
```
majewsky, line 169: Does `json.Encoder.Encode()` add a trailing newline like we used to do? It's not strictly necessary for API compatibility, but makes errors look nicer on the CLI when using curl or similar.

internal/keppel/errors_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+	req := httptest.NewRequest(http.MethodGet, "/v2/test1/foo/manifests/latest", http.NoBody)
+
+	b.ResetTimer()
+	for b.Loop() {
+		rec := httptest.NewRecorder()
+		errObj.WriteAsRegistryV2ResponseTo(rec, req)
+		if rec.Code == 0 {
+			b.Fatal("no status code written")
+		}
```
majewsky, line 29: Please add:
```go
if !bytes.HasPrefix(rec.Body.Bytes(), []byte{'\n'}) {
  b.Fatal("error message should be written with trailing newline to make its appearance in curl etc. look nice")
}
```

internal/test/setup.go
```diff
@@ -313,9 +313,9 @@ func NewSetup(t *testing.T, opts ...SetupOption) Setup {
 
 	// setup initial accounts/repos
 	quotasSetFor := make(map[string]bool)
-	for _, account := range params.Accounts {
-		must.SucceedT(t, s.DB.Insert(account))
-		fd.RecordExistingAccount(s.Ctx, *account, s.Clock.Now()) //nolint:errcheck
+	for i, account := range params.Accounts {
+		must.SucceedT(t, s.DB.Insert(&params.Accounts[i]))
+		fd.RecordExistingAccount(s.Ctx, account, s.Clock.Now()) //nolint:errcheck
```
majewsky:
```suggestion
		must.SucceedT(t, fd.RecordExistingAccount(s.Ctx, account, s.Clock.Now()))
```

majewsky (changes requested):

internal/keppel/errors.go
SuperSandro2000: https://github.com/sapcc/keppel/pull/698#discussion_r3103265553

https://pkg.go.dev/encoding/json#Encoder.Encode

majewsky (approved):

PR #699, opened 2026-04-17
Use Reader for ReadTrivyReport 

SuperSandro2000:


internal/test/helper_validate.go
```diff
@@ -75,16 +75,18 @@ func (s Setup) ExpectManifestsExistInStorage(t *testing.T, repoName string, mani
...
 			t.Errorf("expected manifest %s to exist in the storage, but got: %s", manifest.Digest, err.Error())
 			continue
 		}
+		defer manifestReader.Close()
 		err = manifest.Digest.Validate()
 		if err != nil {
 			t.Errorf("manifest digest %q is not a digest: %s", manifest.Digest, err.Error())
 			continue
 		}
+		manifestBytes := must.ReturnT(io.ReadAll(manifestReader))(t)
```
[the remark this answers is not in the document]
SuperSandro2000: We test like 5 manifests at most and it is a test helper.

internal/keppel/storage_driver.go
```diff
@@ -60,10 +60,13 @@ type StorageDriver interface {
 	// instead.
 	DeleteBlob(ctx context.Context, account models.ReducedAccount, storageID string) error
 
-	ReadManifest(ctx context.Context, account models.ReducedAccount, repoName string, digest digest.Digest) ([]byte, error)
-	WriteManifest(ctx context.Context, account models.ReducedAccount, repoName string, digest digest.Digest, contents []byte) error
+	// On success, the caller owns the returned contents stream and must call Close() on it.
+	ReadManifest(ctx context.Context, account models.ReducedAccount, repoName string, digest digest.Digest) (io.ReadCloser, error)
+	WriteManifest(ctx context.Context, account models.ReducedAccount, repoName string, digest digest.Digest, contents io.Reader) error
```
majewsky: This is the same change previously proposed in #606, and I'm not convinced that it actually does us a whole lot of good. On PUT, we need to buffer into `[]byte` anyway because the processor needs to execute multiple operations (digest checking, parsing, uploading to StorageDriver, uploading to DB) in a way that precludes just using TeeReader for everything. On GET, we usually fetch the manifest from the `manifest_contents` DB table, so there is no avoiding that `[]byte` allocation either.

So we are getting a whole lot more complication in the API in return for optimizing a 2 KB read that we usually cannot optimize anyway.
```suggestion
	// NOTE: We previously explored making these functions pass manifest contents as io.Reader,
	// but adding this extra complication to the interface is not worth it because most operations
	// involving manifests need to buffer into []byte at some point anyway.
	ReadManifest(ctx context.Context, account models.ReducedAccount, repoName string, digest digest.Digest) ([]byte, error)
	WriteManifest(ctx context.Context, account models.ReducedAccount, repoName string, digest digest.Digest, contents []byte) error
```

I would like to see an updated version of this PR that just touches `ReadTrivyReport()`, in order to judge the impact of doing just that.

majewsky (changes requested):

internal/drivers/filesystem/storage.go
```diff
@@ -152,9 +152,13 @@ func (d *StorageDriver) DeleteManifest(ctx context.Context, account models.Reduc
...
 // ReadTrivyReport implements the keppel.StorageDriver interface.
-func (d *StorageDriver) ReadTrivyReport(ctx context.Context, account models.ReducedAccount, repoName string, manifestDigest digest.Digest, format string) ([]byte, error) {
+func (d *StorageDriver) ReadTrivyReport(ctx context.Context, account models.ReducedAccount, repoName string, manifestDigest digest.Digest, format string) (io.ReadCloser, error) {
 	path := d.getTrivyReportPath(account, repoName, manifestDigest, format)
-	return os.ReadFile(path)
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	return f, nil
```
majewsky:
```suggestion
	return os.Open(path)
```

internal/client/download.go
```diff
@@ -43,7 +43,7 @@ type DownloadManifestOpts struct {
 
 // DownloadManifest fetches a manifest from this repository. If an error is
 // returned, it's usually a *keppel.RegistryV2Error.
-func (c *RepoClient) DownloadManifest(ctx context.Context, reference models.ManifestReference, opts *DownloadManifestOpts) (contents []byte, mediaType string, returnErr error) {
+func (c *RepoClient) DownloadManifest(ctx context.Context, reference models.ManifestReference, opts *DownloadManifestOpts) (contents io.ReadCloser, mediaType string, returnErr error) {
```
majewsky: In this PR, there are three locations where DownloadManifest is used in non-test code:

- `Processor.downloadManifestViaInboundCache()` does a ReadAll immediately.
- `RepoClient.doValidateManifest()` does a ReadAll immediately.
- `API.handleDelegatedPullManifest()` does not do a ReadAll, but it loses functionality because it cannot emit a Content-Length header anymore.

I was going to ask for that to be added back by adding an additional `contentLengthHeader string` return argument, but at this point, we are once again adding significant complications for marginal gains, so my verdict is that this function signature change is also not worth it.

cmd/trivyproxy/main.go
```diff
@@ -109,17 +110,17 @@ func (a *API) proxyToTrivy(w http.ResponseWriter, r *http.Request) {
...
+		cleanedErr := strings.ReplaceAll(strings.TrimSpace(stderr.String()), "\n", " ")
 		http.Error(w, fmt.Sprintf("trivy: %s: %s", err, cleanedErr), http.StatusInternalServerError)
 		return
 	}
 
-	w.WriteHeader(http.StatusOK)
 	w.Header().Set("Content-Type", "application/json")
-	w.Write(stdout)
+	w.WriteHeader(http.StatusOK)
+	_, _ = io.Copy(w, &stdout) //nolint:errcheck // nothing we can do about it at this point
```
majewsky:
```suggestion
	_, err := io.Copy(w, &stdout)
	if err != nil {
		logg.Error("could not write Trivy report to stdout: %s", err.Error())
	}
```

internal/api/keppel/manifests.go
```diff
@@ -359,9 +359,11 @@ func (a *API) handleGetTrivyReport(w http.ResponseWriter, r *http.Request) {
 		if respondwith.ObfuscatedErrorText(w, err) {
 			return
 		}
+		defer buf.Close()
+
 		w.Header().Set("Content-Type", "application/json")
 		w.WriteHeader(http.StatusOK)
-		w.Write(buf)
+		io.Copy(w, buf) //nolint:errcheck // we could only log that the client closed the connection early
```
majewsky: Then let's do that:
```suggestion
		_, err := io.Copy(w, buf)
		if err != nil {
			logg.Error("IO error while writing Trivy report: %s", err.Error())
		}
```
Maybe also include which manifest and report format the user was asking for.

majewsky (changes requested):

majewsky (approved):

PR #705, opened 2026-04-22
Remove returned pointers from database_helpers.go

SuperSandro2000:


internal/keppel/database_helpers.go
```diff
@@ -15,7 +15,6 @@ import (
 )
 
 // TODO: rework all functions that may return nil to instead return sql.ErrNoRows
```
wagnerd3: I think all functions which select return directly from the result of `SelectOne`, so this is done and can be deleted, too?

wagnerd3 (dismissed):
I only reviewed superficially as I think all the changes of the signatures should have led to compiler errors so that no spots of usages of those functions should have been missed, if all compiler errors were processed properly ^^

internal/keppel/database_helpers.go
majewsky: Correct, will do.

internal/api/registry/manifests.go
```diff
@@ -171,7 +171,7 @@ func (a *API) handleGetOrHeadManifest(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Length", strconv.FormatUint(uint64(len(manifestBytes)), 10))
 	w.Header().Set("Content-Type", dbManifest.MediaType)
 	w.Header().Set("Docker-Content-Digest", dbManifest.Digest.String())
-	if securityInfo != nil {
+	if !errors.Is(err, sql.ErrNoRows) {
```
majewsky: This particular `err` is from one paragraph ago. If additional loads are introduced, there is a risk that this will be overwritten accidentally. I will amend this to clarify what this check refers to.

SuperSandro2000:
Thanks @majewsky, good changes!

VoigtS (approved):

PR #706, opened 2026-04-22
Remove more return pointers

SuperSandro2000:
Builds on top of https://github.com/sapcc/keppel/pull/705


internal/keppel/config.go
```diff
@@ -57,20 +57,7 @@ func ParseIssuerKey(in string) (crypto.PrivateKey, error) {
...
-	ed25519Key, err1 := jwt.ParseEdPrivateKeyFromPEM(buf)
-	if err1 == nil {
-		return ed25519Key, nil
-	}
-	rsaKey, err2 := jwt.ParseRSAPrivateKeyFromPEM(buf)
-	if err2 == nil {
-		return rsaKey, nil
-	}
-	return nil, fmt.Errorf("neither an ed25519 private key (%q) nor an RSA private key (%q)", err1.Error(), err2.Error())
+	return jwt.ParseEdPrivateKeyFromPEM(buf)
```
wagnerd3, line 60: What about the code in `internal/auth/token.go` for anything != ed25519 - can that be removed too?

internal/test/setup.go
```diff
@@ -250,7 +250,7 @@ func NewSetup(t testing.TB, opts ...SetupOption) Setup {
 		s.Config.AnycastAPIPublicHostname = "registry-global.example.org"
 
 		if params.WithPreviousIssuerKey {
-			key := must.ReturnT(keppel.ParseIssuerKey(UnitTestAnycastIssuerRSAPrivateKey))(t)
+			key := must.ReturnT(keppel.ParseIssuerKey(UnitTestIssuerEd25519PrivateKey))(t)
```
wagnerd3: This is used in line 220, too. So if you have `WithAnycast` and `WithPreviousIssuerKey` in a test, we could have a key valid under both issuers? I can imagine someone blundering into that trap at some point ^^

wagnerd3 (dismissed):

internal/keppel/config.go
SuperSandro2000: Uh, probably, good catch

internal/keppel/config.go
```diff
@@ -57,20 +57,7 @@ func ParseIssuerKey(in string) (crypto.PrivateKey, error) {
...
 }
```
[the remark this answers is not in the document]
majewsky: I don't think we need to complicate the error handling, but we should state the algorithm requirement in the docs.

majewsky (changes requested):

majewsky (dismissed):

majewsky (approved):

PR #710, opened 2026-05-05
Add new capacity resource

SuperSandro2000:


internal/api/registry/uploads.go
```diff
@@ -87,13 +87,23 @@ func (a *API) handleStartBlobUpload(w http.ResponseWriter, r *http.Request) {
...
+		bytesUsage, err := a.sd.UsedBytes(r.Context(), account.AuthTenantID)
+		if respondWithError(w, r, err) {
+			return
+		}
+		if bytesUsage >= quotas.Bytes {
+			msg := fmt.Sprintf("bytes quota exceeded (quota = %d, usage = %d)", quotas.Bytes, bytesUsage)
+			keppel.ErrDenied.With(msg).WithStatus(http.StatusConflict).WriteAsRegistryV2ResponseTo(w, r)
+			return
+		}
+	}
```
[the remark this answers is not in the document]
SuperSandro2000: We will consider this after some real world tests.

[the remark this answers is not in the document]
SuperSandro2000: This is intentional to avoid clients aborting and than immediately retrying as this is not a hard but a soft limit.

internal/api/liquid/api.go
```diff
@@ -122,4 +145,15 @@ func liquidConvertQuotaResponse(resp processor.QuotaResponse) liquid.ServiceUsag
 			},
 		},
 	}
+
+	if res, ok := resp.Bytes.Unpack(); ok {
+		su.Resources["capacity"] = &liquid.ResourceUsageReport{
+			Quota: Some(int64(res.Quota)), //nolint:gosec // quota is admin controlled
```
wagnerd3: Ahhhh I think that would be a problem, if `capacity` quota was admin controlled. I think we definitely need to add an autogrowing quota for `capacity`, because otherwise the customers can indefinitely grow the centrally managed storage account. 

Even if this is not part of this PR, I would say either a) we should have a TODO here to change this when we introduce proper quota or b) directly safeguard the casting.

internal/api/keppel/quotas_test.go
```diff
@@ -6,65 +6,122 @@ package keppelv1_test
...
 	"github.com/sapcc/go-bits/httptest"
-	"github.com/sapcc/go-bits/must"
 	"go.xyrillian.de/gg/jsonmatch"
-	. "go.xyrillian.de/gg/option"
 
 	"github.com/sapcc/keppel/internal/models"
 	"github.com/sapcc/keppel/internal/test"
 )
 
 func TestQuotasAPI(t *testing.T) {
```
wagnerd3, line 18: Should we make this 
```suggestion
func TestQuotasAPIWithoutCapacity(t *testing.T) {
```

internal/api/keppel/quotas_test.go
```diff
@@ -196,18 +243,239 @@ func TestQuotasAPI(t *testing.T) {
...
 		withPerms("changequota:tenant1"),
 		httptest.WithJSONBody(map[string]any{
-			"manifests": map[string]any{"quota": 100, "usage": 10},
+			"manifests": map[string]any{"quota": 5},
+		}),
+	).ExpectText(t, http.StatusUnprocessableEntity, "requested manifest quota (5) is below usage (10)\n")
+	s.Auditor.ExpectEvents(t /*, nothing */)
+}
+
+func TestQuotasAPIWithBytes(t *testing.T) {
```
wagnerd3, line 261: I would try to refer to it what it is from API side and not from internal side here.
```suggestion
func TestQuotasAPIWithCapacity(t *testing.T) {
```

internal/test/setup.go
```diff
@@ -97,6 +98,11 @@ func WithQuotas(params *setupParams) {
 	params.WithQuotas = true
 }
 
+// WithBytesQuotas is a SetupOption that enables bytes quota.
+func WithBytesQuotas(params *setupParams) {
```
wagnerd3, line 109: Same here and everywhere else, I would refer to it as 
```suggestion
func WithCapacityQuotas(params *setupParams) {
```

internal/api/liquid/api.go
```diff
@@ -34,7 +35,19 @@ func (a *API) handleLiquidGetInfo(w http.ResponseWriter, r *http.Request) {
 				HasQuota:    true,
 			},
 		},
-	})
+	}
+
+	if a.cfg.TrackBytesQuota {
```
wagnerd3, line 89: We should stick to the naming from API consumer view here. 
```suggestion
	if a.cfg.TrackCapacityQuota {
```

wagnerd3 (dismissed):
The point about the naming applies in many cases, I stopped commenting now. Though I understand that my POV is very billing-focused, I can see that a keppel-developer maybe understands the `bytes` naming better? Probably 2nd opinion from @majewsky would help.

internal/test/setup.go
majewsky: Regarding the naming, I suggested `bytes` for the quota because that follows the established naming conventions within OpenStack. Using `capacity` there would be problematic because quotas are only structured as `map[string](u)int64` throughout OpenStack and thus cannot indicate their unit. As a point of comparison, Cinder has quotas called `volumes`, `snapshots` and `gigabytes` (with Limes renaming the latter to `capacity` because it has a better way to represent the unit information).

wagnerd3: Then disregard this point. I just know the outside view so this makes total sense to me.

internal/api/liquid/api.go
majewsky: > I think we definitely need to add an autogrowing quota for capacity

Yes, that is the intent. The policy is already such that only `rule:resource_service` (i.e. Limes itself) can set quotas.

internal/models/quotas.go
```diff
@@ -8,15 +8,19 @@ package models
 // The JSON serialization is used in audit events for quota changes.
 type Quotas struct {
 	AuthTenantID  string `db:"auth_tenant_id" json:"-"`
+	Bytes         uint64 `db:"bytes" json:"bytes,omitempty"`
 	ManifestCount uint64 `db:"manifests" json:"manifests"`
 }
```
majewsky, line 13: Can we get rid of these `json:"..."` tags? As in, does it break any tests if we do? I'm assuming that this used to be relevant because those records were directly rendered on the quota GET endpoint, but now there is the indirection of `processor.QuotaResponse`.

internal/models/quotas.go
```diff
@@ -8,15 +8,19 @@ package models
...
 	// Right now, the default quota is always 0. The value of having this function
 	// is to ensure that we only need to change this place if this ever changes.
-	return Quotas{
+	quota := Quotas{
 		AuthTenantID:  authTenantID,
+		Bytes:         0,
 		ManifestCount: 0,
 	}
+
+	return quota
```
majewsky: Can be shortened back to `return Quotas{...}`.

internal/processor/quotas.go
```diff
@@ -31,7 +34,8 @@ type SingleQuotaResponse struct {
 
 // QuotaRequest is the request body payload for PUT /keppel/v1/quotas/:auth_tenant_id.
 type QuotaRequest struct {
-	Manifests SingleQuotaRequest `json:"manifests"`
+	Bytes     Option[SingleQuotaRequest] `json:"bytes,omitzero"`
+	Manifests Option[SingleQuotaRequest] `json:"manifests,omitzero"`
```
majewsky:
```suggestion
	// This field is always required. Option[] is only used to distinguish a quota set to 0 from a missing quota.
	Manifests Option[SingleQuotaRequest] `json:"manifests,omitzero"`
```

internal/api/keppel/quotas_test.go
```diff
@@ -196,18 +243,239 @@ func TestQuotasAPI(t *testing.T) {
...
+	// NOTE: This tests both the Keppel-native quota API and the LIQUID API which accesses the same logic.
```
majewsky:
```suggestion
	// NOTE: This tests both the Keppel-native quota API and the LIQUID API which access the same logic.
```

majewsky (changes requested):

internal/api/liquid/api.go
wagnerd3: Ah, this is meant by admin, then I am fine with that.

SuperSandro2000:
> The point about the naming applies in many cases, I stopped commenting now. Though I understand that my POV is very billing-focused, I can see that a keppel-developer maybe understands the `bytes` naming better? Probably 2nd opinion from @majewsky would help.

See https://github.wdf.sap.corp/sap-cloud-infrastructure/storage-resource-services/issues/657 for the naming. I probably mixed some cases up though 😅

internal/models/quotas.go
SuperSandro2000: The AuditQuotas type is still using this type and it breaks test if removed.

```
--- FAIL: TestQuotasAPIWithBytes (0.18s)
    quotas_test.go:358: value mismatch at /events/0/target/attachments/0/content: expected "{\"manifests\":0}", but got "{\"AuthTenantID\":\"tenant1\",\"Bytes\":0,\"ManifestCount\":0}"
    quotas_test.go:358: value mismatch at /events/0/target/attachments/1/content: expected "{\"bytes\":5000000,\"manifests\":50}", but got "{\"AuthTenantID\":\"tenant1\",\"Bytes\":5000000,\"ManifestCount\":50}"
    quotas_test.go:400: value mismatch at /events/0/target/attachments/0/content: expected "{\"bytes\":5000000,\"manifests\":50}", but got "{\"AuthTenantID\":\"tenant1\",\"Bytes\":5000000,\"ManifestCount\":50}"
    quotas_test.go:400: value mismatch at /events/0/target/attachments/1/content: expected "{\"bytes\":50000000,\"manifests\":100}", but got "{\"AuthTenantID\":\"tenant1\",\"Bytes\":50000000,\"ManifestCount\":100}"
2026/05/08 17:25:38 REQUEST:  - - "GET https://trivy.example.org/trivy?format=spdx-json&image=registry.example.org%2Ftest1%2Ffoo%40sha256%3Ae3c1e46560a7ce30e3d107791e1f60a588eda9554564a5d17aa365e53dd6ae58 HTTP/1.1" 200 24970 "-" "-" 0.002s
--- FAIL: TestQuotasAPI (0.11s)
    quotas_test.go:138: value mismatch at /events/0/target/attachments/0/content: expected "{\"manifests\":0}", but got "{\"AuthTenantID\":\"tenant1\",\"Bytes\":0,\"ManifestCount\":0}"
    quotas_test.go:138: value mismatch at /events/0/target/attachments/1/content: expected "{\"manifests\":50}", but got "{\"AuthTenantID\":\"tenant1\",\"Bytes\":0,\"ManifestCount\":50}"
    quotas_test.go:179: value mismatch at /events/0/target/attachments/0/content: expected "{\"manifests\":50}", but got "{\"AuthTenantID\":\"tenant1\",\"Bytes\":0,\"ManifestCount\":50}"
    quotas_test.go:179: value mismatch at /events/0/target/attachments/1/content: expected "{\"manifests\":100}", but got "{\"AuthTenantID\":\"tenant1\",\"Bytes\":0,\"ManifestCount\":100}"
FAIL
```

internal/api/registry/uploads.go
```diff
@@ -87,13 +87,23 @@ func (a *API) handleStartBlobUpload(w http.ResponseWriter, r *http.Request) {
...
+	if a.cfg.TrackBytesQuota {
+		bytesUsage, err := a.sd.UsedBytes(r.Context(), account.AuthTenantID)
+		if respondWithError(w, r, err) {
+			return
+		}
+		if bytesUsage >= quotas.Bytes {
+			msg := fmt.Sprintf("bytes quota exceeded (quota = %d, usage = %d)", quotas.Bytes, bytesUsage)
+			keppel.ErrDenied.With(msg).WithStatus(http.StatusConflict).WriteAsRegistryV2ResponseTo(w, r)
+			return
+		}
```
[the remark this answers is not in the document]
SuperSandro2000: @majewsky how do we want to do the activation? If we enable the feature, limes will not instantly be able to set it everywhere. Do we do the migration via a SQL query?

internal/api/liquid/api.go
```diff
@@ -17,14 +18,14 @@ import (
 	"github.com/sapcc/keppel/internal/processor"
 )
 
-// Increment this whenever the output of handleLiquidGetInfo() changes.
-const LiquidInfoVersion int64 = 2
+var liquidInfoVersion int64 = time.Now().Unix()
```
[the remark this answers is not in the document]
SuperSandro2000: @wagnerd3 @majewsky I do not know if that is a concern if we spawn multiple API pods.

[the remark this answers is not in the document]
majewsky: Oh yeah, that is indeed a problem. We will have to think about this. The obvious solution would be to deploy the LIQUID API as a standalone process that does not allow for horizontal scaling, but maybe we can come up with something that does not increase deployment complexity.

internal/api/registry/uploads.go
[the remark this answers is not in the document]
majewsky: OpenStack often uses a default of `-1` to mean `unlimited`, and then in practice Limes overrides that as soon as it assigns quota. This would allow for a non-disruptive rollout.

Since I don't want to go into the negative-numbers territory, we could default to `math.MaxUint64` (or maybe `math.MaxInt64` to not overflow in case of int64 conversions).

internal/api/liquid/api.go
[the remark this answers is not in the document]
majewsky: Yeah, we will do a separate liquid-keppel pod. Please add a new subcommand `keppel server liquidapi` in `cmd/liquidapi`, remove the LIQUID API from `keppel server api` and adjust the Helm chart to deploy one replica of the new API (same as for the liquids in the Limes Helm chart).

internal/test/setup.go
```diff
@@ -301,6 +317,9 @@ func NewSetup(t testing.TB, opts ...SetupOption) Setup {
 	if params.WithKeppelAPI {
 		apis = append(apis, keppelv1.NewAPI(s.Config, ad, fd, sd, icd, s.DB, s.Auditor, params.RateLimitEngine).OverrideTimeNow(s.Clock.Now))
 	}
+	if params.WithLiquidAPI {
+		apis = append(apis, keppelv1.NewLiquidAPI(s.Config, ad, sd, s.DB, s.Auditor) /*.OverrideTimeNow(s.Clock.Now)*/)
+	}
```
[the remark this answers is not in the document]
majewsky: Since we don't rely on overriding `timeNow` in tests, we can replace that dependency injection hook and just use `time.Now` directly in the implementation.

internal/api/liquid/api.go
```diff
@@ -5,26 +5,73 @@ package keppelv1
...
 	. "go.xyrillian.de/gg/option"
 
+	"github.com/sapcc/keppel/internal/auth"
 	"github.com/sapcc/keppel/internal/keppel"
 	"github.com/sapcc/keppel/internal/processor"
 )
 
-// Increment this whenever the output of handleLiquidGetInfo() changes.
-const LiquidInfoVersion int64 = 2
+var liquidInfoVersion int64 = time.Now().Unix()
```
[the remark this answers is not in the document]
majewsky: Sandro already knows this, but to document this decision for Copilot et al: No. We are explicitly moving away from hardcoding a version number because the ServiceInfo can now change based on configuration. And having it change on restart is fine because a ServiceInfo version bump without changes to the ServiceInfo contents is practically a no-op in Limes.

cmd/api/main.go
```diff
@@ -107,20 +106,6 @@ func run(cmd *cobra.Command, args []string) {
...
-	}
-	logg.Debug("initializing Redis connection...")
-
-	opts, err := keppel.GetRedisOptions("KEPPEL_REDIS")
-	if err != nil {
-		return nil, fmt.Errorf("cannot parse Redis URL: %s", err.Error())
-	}
-	return redis.NewClient(opts), nil
-}
-
```
majewsky, line 123: This can stay here. We don't need rate-limiting in the LIQUID API; it's cloud-admin-only anyway.

internal/api/liquid/api.go
majewsky, line 1: Since this now becomes its own API struct that is separate from the Keppel API, please move it to its own package for clarity -> `internal/api/liquid/api.go`

majewsky (changes requested):
Just a few broad comments. Since Copilot flagged a lot of stuff, I will wait with my review until those comments were addressed.

cmd/liquidapi/main.go
```diff
@@ -0,0 +1,82 @@
...
+
+	ad := must.Return(keppel.NewAuthDriver(ctx, osext.MustGetenv("KEPPEL_DRIVER_AUTH"), nil))
+	sd := must.Return(keppel.NewStorageDriver(osext.MustGetenv("KEPPEL_DRIVER_STORAGE"), ad, cfg))
+
+	// wire up HTTP handlers
+	corsMiddleware := cors.New(cors.Options{
+		AllowedOrigins: []string{"*"},
+		AllowedMethods: []string{"HEAD", "GET", "POST", "PUT", "DELETE"},
+		AllowedHeaders: []string{"Content-Type", "User-Agent", "Authorization", "X-Auth-Token", keppelv1.SubleaseHeader},
+	})
```
majewsky: We don't need a CORS middleware for this. This is not an API that's expected to be consumable by regular users in the web dashboard.

cmd/liquidapi/main.go
```diff
@@ -0,0 +1,82 @@
...
+		httpapi.HealthCheckAPI{
+			SkipRequestLog: true,
+			Check: func() error {
+				return db.Db.PingContext(ctx)
+			},
+		},
+		httpapi.WithGlobalMiddleware(corsMiddleware.Handler),
+		pprofapi.API{IsAuthorized: pprofapi.IsRequestFromLocalhost},
+		// This needs to be at the end because it is the fallback match for all
+		// paths that are not otherwise defined.
```
majewsky: The thing that this comment refers to (the `guiRedirecter` API) does not exist here.
```suggestion
```

docs/operator-guide.md
majewsky, line 136: New entry here please:
```
- optionally, one instance of `keppel server liquidapi` to enable integration with [Limes](https://github.com/sapcc/limes) if desired,
```

docs/operator-guide.md
```diff
@@ -268,6 +269,15 @@ downloaded and validated every 30 seconds. The result of the test is published a
 the test fails, a detailed error message is logged in stderr. If the setup phase fails, an error message is logged as
 well and the program immediately exits with non-zero status.
 
+### Liquid configuration options
+
+These options are only useful when the Limes LIQUID API is deployed.
```
majewsky:
```suggestion
These options are understood by `keppel server liquidapi`:
```

In general, this entire series of a billion tables with "these env vars are for $SET_OF_PROCESSES" is getting messy. Can you please try prompting the robot to rewrite it into a single table with

```markdown
| Name | Default | Relevant for components | Description |
```

and see if that looks like anything useful when rendered?

docs/operator-guide.md
```diff
@@ -276,6 +286,7 @@ These options are only useful when the Trivy proxy is deployed but the Keppel AP
 | -------- | ------- | ----------- |
 | `KEPPEL_TRIVY_ADDITIONAL_PULLABLE_REPOS` | *(optional)* | It adds additional scopes to the token issued by the API and the janitor which is meant to allow the trivy components to pull their DB OCI images from the respective repos. |
 | `KEPPEL_TRIVY_DB_MIRROR_PREFIX` | *(required)* | Prefix under which trivy can find its database. This might be a mirror or ghcr.io. |
+| `KEPPEL_TRIVY_LISTEN_ADDRESS` | :8080 | Listen address for HTTP server. |
```
majewsky: This option is misplaced here. These variables are for the API and janitor, not for the trivyproxy.

internal/api/keppel/quotas_test.go
```diff
@@ -6,65 +6,123 @@ package keppelv1_test
...
+	s.RespondTo(ctx, "PUT /liquid/v1/projects/tenant1/quota",
+		withPerms("changequota:tenant1"),
+		httptest.WithJSONBody(map[string]any{
+			"resources": map[string]any{
+				"capacity": map[string]any{"quota": 100},
+			},
+		}),
+	).ExpectBody(t, http.StatusUnprocessableEntity, []byte("request does not contain manifest quota\n"))
+
+	// basic error cases: trying to set bytes quota but it is not enabled
```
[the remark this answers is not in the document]
majewsky: We can probably just get rid of this line here.

internal/models/quotas.go
```diff
@@ -3,20 +3,28 @@
...
+	Bytes         uint64 `db:"bytes" json:"bytes,omitempty"`
 	ManifestCount uint64 `db:"manifests" json:"manifests"`
 }
 
 // DefaultQuotas creates a new Quotas instance with the default quotas.
-func DefaultQuotas(authTenantID string) Quotas {
-	// Right now, the default quota is always 0. The value of having this function
-	// is to ensure that we only need to change this place if this ever changes.
-	return Quotas{
+func DefaultQuotas(authTenantID string, trackBytesQuota bool) Quotas {
```
majewsky: Since we have an intuitive option for a structured type, I would like to prefer that over a `bool` argument:
```suggestion
func DefaultQuotas(authTenantID string, cfg keppel.Configuration) Quotas {
```

internal/models/quotas.go
```diff
@@ -3,20 +3,28 @@
...
+	quotas := Quotas{
 		AuthTenantID:  authTenantID,
+		Bytes:         0,
 		ManifestCount: 0,
 	}
+
+	if !trackBytesQuota {
+		quotas.Bytes = math.MaxInt64
+	}
```
majewsky: After thinking about this again, I would like to not go with `MaxInt64`. It does make the most logical sense, but Limes contains places where quota is summed over multiple projects, and `MaxInt64` is going to blow up there (in the arithmetic overflow sense). We should keep to the OpenStack standard, as messy as it is, and represent "unlimited" as `-1` (i.e. the type for `quotas.Bytes` should be `int64`), because Limes has proper handling for this case. When enforcing quotas, if a negative quota value is found, then the respective quota check shall be skipped.

internal/keppel/database.go
```diff
@@ -427,6 +427,14 @@ var sqlMigrations = map[string]string{
 			ALTER COLUMN next_check_at SET NOT NULL,
 			DROP CONSTRAINT next_check_at_only_null_when_rotten;
 	`,
+	"054_add_quotas_bytes.up.sql": `
+		ALTER TABLE quotas
+			ADD COLUMN bytes BIGINT NOT NULL DEFAULT 0;
```
majewsky: Consequently this should have a default of `-1` then.

majewsky (changes requested):

majewsky (changes requested):
I forgot to write the overall review notes: As much as I still have a bunch of notes, we are getting to the finish line. Please prepare the helm-charts PR that adds the liquidapi deployment (1 replica with strategy `Recreate`) and service/ingress, and adjusts the catalog-seed to point to that new service/ingress.

docs/operator-guide.md
```diff
@@ -166,54 +165,61 @@ your environment:
...
+| `KEPPEL_API_ANYCAST_FQDN` | *(optional)* | api | Full domain name where users reach any keppel-api in this peer group, usually via anycast. Requests for accounts not held locally are reverse-proxied to the correct peer. Limited to anonymous authorization; cannot be used for pushing. |
+| `KEPPEL_API_LISTEN_ADDRESS` | `:8080` | api | Listen address for the HTTP server. |
+| `KEPPEL_API_PUBLIC_FQDN` | *(required)* | api, janitor | Full domain name where users reach keppel-api. |
+| `KEPPEL_AUDIT_RABBITMQ_HOSTNAME` | `localhost` | api, janitor | Hostname of the RabbitMQ server. |
+| `KEPPEL_AUDIT_RABBITMQ_PASSWORD` | `guest` | api, janitor | Password for the RabbitMQ user. |
+| `KEPPEL_AUDIT_RABBITMQ_PORT` | `5672` | api, janitor | Port number for the RabbitMQ connection. |
+| `KEPPEL_AUDIT_RABBITMQ_QUEUE_NAME` | *(required to enable audit trail)* | api, janitor | Name of the queue that will hold audit events, published to the default exchange. If not set, audit events are only written to the debug log. |
+| `KEPPEL_AUDIT_RABBITMQ_USERNAME` | `guest` | api, janitor | RabbitMQ username. |
```
majewsky: liquidapi also writes audit events.

docs/operator-guide.md
```diff
@@ -166,54 +165,61 @@ your environment:
...
+| `KEPPEL_DB_CONNECTION_OPTIONS` | *(optional)* | api, janitor | Additional database connection options. |
+| `KEPPEL_DB_HOSTNAME` | `localhost` | api, janitor | Hostname of the PostgreSQL server. |
+| `KEPPEL_DB_NAME` | `keppel` | api, janitor | Name of the database. |
+| `KEPPEL_DB_PASSWORD` | *(optional)* | api, janitor | Password for the database user. |
+| `KEPPEL_DB_PORT` | `5432` | api, janitor | Port on which the PostgreSQL service is running. |
+| `KEPPEL_DB_USERNAME` | `postgres` | api, janitor | Username for the database connection. |
```
majewsky: liquidapi also connects to the DB.

majewsky (changes requested):
I looked over the current state of the code change and had no new notes. I also resolved conversations that were addressed.

Doc changes look good, except for a few small nitpicks:

SuperSandro2000:
> I forgot to write the overall review notes: As much as I still have a bunch of notes, we are getting to the finish line. Please prepare the helm-charts PR that adds the liquidapi deployment (1 replica with strategy `Recreate`) and service/ingress, and adjusts the catalog-seed to point to that new service/ingress.

That still needs to be done.

SuperSandro2000:
> I forgot to write the overall review notes: As much as I still have a bunch of notes, we are getting to the finish line. Please prepare the helm-charts PR that adds the liquidapi deployment (1 replica with strategy `Recreate`) and service/ingress, and adjusts the catalog-seed to point to that new service/ingress.

https://github.com/sapcc/helm-charts/pull/11880

majewsky (approved):

PR #721, opened 2026-05-18
remove last usage of assert.HTTPResponseBody

majewsky:
This concludes the assert.HTTPRequest removal refactor.

### DIFF
--- a/internal/tasks/manifests_test.go
+++ b/internal/tasks/manifests_test.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"os"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


internal/tasks/manifests_test.go
```diff
@@ -616,9 +617,9 @@ func TestCheckTrivySecurityStatus(t *testing.T) {
 			images[0].Layers[0].Digest, images[1].Layers[0].Digest, images[2].Layers[0].Digest, images[3].Layers[0].Digest, images[4].Layers[0].Digest, images[5].Layers[0].Digest)
 
 		// for scannable images, a report should now be cached in storage
-		s.ExpectTrivyReportExistsInStorage(t, imageManifests[0], "json", assert.JSONFixtureFile("fixtures/trivy/report-vulnerable.json"))
-		s.ExpectTrivyReportExistsInStorage(t, imageManifests[1], "json", assert.JSONFixtureFile("fixtures/trivy/report-clean.json"))
-		s.ExpectTrivyReportExistsInStorage(t, imageManifests[2], "json", assert.JSONFixtureFile("fixtures/trivy/report-vulnerable.json"))
+		s.ExpectTrivyReportExistsInStorage(t, imageManifests[0], "json", must.ReturnT(os.ReadFile("fixtures/trivy/report-vulnerable.json"))(t))
```
wagnerd3, line 620: I thought it would have been nice to use `JQModifiableJSONFixture` consistently, even when we don't want to mod them. But this is fine, too. What do you think?

wagnerd3 (approved):

internal/tasks/manifests_test.go
majewsky: This would require making the respective argument of `ExpectTrivyReportExistsInStorage` a `jsonmatch.Diffable`, which would make it harder to expand to non-JSON format types in the future. We agreed to leave it as is for now.

PR #726, opened 2026-05-22
replace Gorp with Oblast

majewsky:
meme

The first commit in this branch adds a benchmark for the most important read-only API operations (GetManifest, GetBlob and ListTags), so we can compare before:

```
goos: linux
goarch: amd64
pkg: github.com/sapcc/keppel/internal/api/registry
cpu: AMD Ryzen 9 7900X 12-Core Processor
BenchmarkImportantReadOperations/GetManifest-24         	    1156	   1020868 ns/op	   32467 B/op	     462 allocs/op
BenchmarkImportantReadOperations/GetBlob-24             	    3912	    305104 ns/op	   27820 B/op	     336 allocs/op
BenchmarkImportantReadOperations/ListTags-24            	    4987	    239501 ns/op	   21724 B/op	     275 allocs/op
PASS
ok  	github.com/sapcc/keppel/internal/api/registry	3.855s
```

And after:

```
goos: linux
goarch: amd64
pkg: github.com/sapcc/keppel/internal/api/registry
cpu: AMD Ryzen 9 7900X 12-Core Processor
BenchmarkImportantReadOperations/GetManifest-24         	    1203	   1001007 ns/op	   30508 B/op	     425 allocs/op
BenchmarkImportantReadOperations/GetBlob-24             	    3854	    299170 ns/op	   27392 B/op	     324 allocs/op
BenchmarkImportantReadOperations/ListTags-24            	    5119	    242539 ns/op	   22430 B/op	     289 allocs/op
PASS
ok  	github.com/sapcc/keppel/internal/api/registry	3.885s
````

All in all, a slight improvement on GetManifest and GetBlob, and a slight regression on ListTags. Not really a decisive victory, but I remain hopeful that the move to pgx that Oblast can enable might still be a big win.

The implementation is mostly straightforward, though I will note one implementation change that I happened across while going through the compiler errors:

### CODE REVIEW

File: internal/api/keppel/repos.go [hunk 1]
```diff
@@ -238,26 +242,28 @@ func (a *API) handleDeleteRepository(w http.ResponseWriter, r *http.Request) {
...
-			http.Error(w, processor.DeleteManifestBlockedByTagPolicyError{Policy: tagPolicy}.Error(), http.StatusConflict)
-			return
+		if tagPolicy.BlockDelete && tagPolicy.MatchesRepository(repo.Name) {
+			for manifestDigest, manifestTags := range tagsByManifestDigest {
+				if tagPolicy.MatchesTags(manifestTags) {
+					err := processor.DeleteManifestBlockedByTagPolicyError{Digest: manifestDigest, Policy: tagPolicy}
+					http.Error(w, err.Error(), http.StatusConflict)
+					return
+				}
+			}
```
majewsky: When considering a repository deletion request, this would evaluate tag policies by looking at all images of all tags at once. This is incorrect, because tag policies can have negative matches, but those should only apply per manifest, not for all manifests of a repo as a whole. If the policy is "block deletion if tagged `.*latest`, but not if tagged `.*testingonly`", the previous logic would allow to delete a repo containing an image tagged `foo-latest` if there is also another unrelated image tagged `bar-testingonly` in the same repo. This is not what we want, so I quickly rewrote this to evaluate tag policies per manifest.

SuperSandro2000: Please do this in an extra commit.

internal/api/registry/uploads.go
```diff
@@ -679,7 +690,7 @@ func (a *API) createBlobFromUpload(ctx context.Context, account models.ReducedAc
 	}
 	defer sqlext.RollbackUnlessCommitted(tx)
 
-	_, err = tx.Delete(&upload)
+	err = models.UploadStore.Delete(ctx, a.db, upload)
```
wagnerd3: I see the `tx` going away from the call - can this be right?

internal/tasks/account_management.go
```diff
@@ -55,7 +56,14 @@ func (j *Janitor) discoverManagedAccount(_ context.Context, _ prometheus.Labels)
 
 	// if there is a managed account that does not exist yet, create it
 	var existingAccountNames []models.AccountName
-	_, err = j.db.Select(&existingAccountNames, "SELECT name FROM accounts WHERE is_managed")
+	err = sqlext.ForeachRow(j.db, `SELECT name FROM accounts WHERE is_managed`, nil, func(rows *sql.Rows) error {
```
wagnerd3: Is there no way to write this in a short form with oblast? (single column statement goes to array of a single column type?) Maybe separate method for this case for better type checking?

internal/keppel/database_helpers.go
```diff
@@ -64,85 +52,69 @@ var blobGetQueryByRepoID = sqlext.SimplifyWhitespace(`
...
-func FindManifest(db gorp.SqlExecutor, repo models.ReducedRepository, manifestDigest digest.Digest) (models.Manifest, error) {
-	var manifest models.Manifest
-	err := db.SelectOne(&manifest,
-		"SELECT * FROM manifests WHERE repo_id = $1 AND digest = $2", repo.ID, manifestDigest)
-	return manifest, err
-}
+var (
+	manifestGetQueryByRepoID   = models.ManifestStore.MustPrepareSelectQueryWhere(`repo_id = $1 AND digest = $2`)
+	manifestGetQueryByRepoName = models.ManifestStore.MustPrepareSelectQueryWhere(
+		`repo_id = (SELECT id FROM repos WHERE account_name = $1 AND name = $2) AND digest = $3`)
```
wagnerd3, line 95: Does subquery provide the same speed as join with filter? I don't know how this behaves in Postgresql usually, I only know from HANA that you need to be careful with such stuff.

wagnerd3 (commented):
I skimmed over it, if you are sufficiently confident, we can merge.

internal/keppel/database_helpers.go
majewsky: The plans are not exactly identical, but performance is comparable:

```
EXPLAIN ANALYZE
 SELECT * FROM manifests
  WHERE repo_id = (SELECT id FROM repos WHERE account_name = 'foo' AND name = 'bar') AND digest = 'sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855';

+----------------------------------------------------------------------------------------------------------------------------------------------+
| QUERY PLAN                                                                                                                                   |
|----------------------------------------------------------------------------------------------------------------------------------------------|
| Index Scan using manifests_pkey on manifests  (cost=2.92..5.14 rows=1 width=849) (actual time=0.010..0.011 rows=0.00 loops=1)                |
|   Index Cond: ((repo_id = (InitPlan 1).col1) AND (digest = 'sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'::text)) |
|   Index Searches: 0                                                                                                                          |
|   Buffers: shared hit=2                                                                                                                      |
|   InitPlan 1                                                                                                                                 |
|     ->  Index Scan using repos_account_name_name_key on repos  (cost=0.28..2.50 rows=1 width=8) (actual time=0.007..0.008 rows=0.00 loops=1) |
|           Index Cond: ((account_name = 'foo'::text) AND (name = 'bar'::text))                                                                |
|           Index Searches: 1                                                                                                                  |
|           Buffers: shared hit=2                                                                                                              |
| Planning:                                                                                                                                    |
|   Buffers: shared hit=29                                                                                                                     |
| Planning Time: 0.216 ms                                                                                                                      |
| Execution Time: 0.025 ms                                                                                                                     |
+----------------------------------------------------------------------------------------------------------------------------------------------+
EXPLAIN 13

 EXPLAIN ANALYZE
 SELECT m.* FROM manifests m JOIN repos r ON m.repo_id = r.id
  WHERE r.account_name = 'foo' AND r.name = 'bar' AND m.digest = 'sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855';

+----------------------------------------------------------------------------------------------------------------------------------------------+
| QUERY PLAN                                                                                                                                   |
|----------------------------------------------------------------------------------------------------------------------------------------------|
| Nested Loop  (cost=0.70..7.36 rows=1 width=849) (actual time=0.008..0.009 rows=0.00 loops=1)                                                 |
|   Buffers: shared hit=2                                                                                                                      |
|   ->  Index Scan using repos_account_name_name_key on repos r  (cost=0.28..2.50 rows=1 width=8) (actual time=0.008..0.008 rows=0.00 loops=1) |
|         Index Cond: ((account_name = 'foo'::text) AND (name = 'bar'::text))                                                                  |
|         Index Searches: 1                                                                                                                    |
|         Buffers: shared hit=2                                                                                                                |
|   ->  Index Scan using manifests_pkey on manifests m  (cost=0.42..2.64 rows=1 width=849) (never executed)                                    |
|         Index Cond: ((repo_id = r.id) AND (digest = 'sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'::text))        |
|         Index Searches: 0                                                                                                                    |
| Planning:                                                                                                                                    |
|   Buffers: shared hit=18                                                                                                                     |
| Planning Time: 0.216 ms                                                                                                                      |
| Execution Time: 0.030 ms                                                                                                                     |
+----------------------------------------------------------------------------------------------------------------------------------------------+
EXPLAIN 13
```

internal/api/registry/uploads.go
majewsky: Copy-paste error. I checked the full diff for other unbalanced additions of `/\.(?i)db/`, but this is the only one.

internal/tasks/account_management.go
majewsky: Not in Oblast, since it is only about loading and storing records. I might expand its API to cover the single-column usecase, but for now, I added `keppel.SelectSeveralValues()` to match `keppel.SelectOneValue()`.

internal/api/keppel/repos.go
majewsky: Done in #738, and thus now expunged from this branch's diff.

wagnerd3 (approved):

PR #732, opened 2026-06-01
include username in request logs

majewsky:
Using the newly added `httpapi.IdentifyUser()` API.

### DIFF
--- a/internal/tasks/janitor.go
+++ b/internal/tasks/janitor.go
@@ -112,7 +112,7 @@ func (uid janitorUserIdentity) UserType() keppel.UserType {
 
 // UserName implements the keppel.UserIdentity interface.
 func (uid janitorUserIdentity) UserName() string {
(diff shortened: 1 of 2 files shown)


wagnerd3 (approved):

internal/tasks/janitor.go
```diff
@@ -112,7 +112,7 @@ func (uid janitorUserIdentity) UserType() keppel.UserType {
 
 // UserName implements the keppel.UserIdentity interface.
 func (uid janitorUserIdentity) UserName() string {
-	return ""
+	return "janitor:" + uid.TaskName
```
majewsky, line 115: FYI: This should never be reached because the janitor user identity is only used for audit logging. But just in case this changes in the future, I'm not leaving this as "empty string" to avoid it being misclassified as an anonymous user.

PR #738, opened 2026-06-15
api/keppel: fix evaluation of tag policies during repo deletion

majewsky:
When considering a repository deletion request, the previous logic evaluated tag policies by looking at all tags of all images at once. This is incorrect, because tag policies can have negative matches, but those should only apply per manifest, not for all manifests of a repo as a whole. If the policy is "block deletion if tagged `.*latest`, but not if tagged `.*testingonly`", the previous logic would allow to delete a repo containing an image tagged `foo-latest` if there is also another unrelated image tagged `bar-testingonly` in the same repo.

### DIFF
--- a/internal/api/keppel/repos.go
+++ b/internal/api/keppel/repos.go
@@ -238,22 +238,25 @@ func (a *API) handleDeleteRepository(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var (
-		tagResults []models.Tag
-		tags       []string
-	)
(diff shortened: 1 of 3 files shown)


SuperSandro2000:
Shouldn't we add a test case for this exact case?

majewsky:
> Shouldn't we add a test case for this exact case?

There already is, as specifically called out in the updated comment on the test.

VoigtS (approved):

SuperSandro2000 (approved):

PR #751, opened 2026-07-18
Add multi storage driver

SuperSandro2000:


internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,415 @@
...
+
+func init() {
+	keppel.StorageDriverRegistry.Add(func() keppel.StorageDriver { return &StorageDriver{} })
+}
+
+// PluginTypeID implements the keppel.StorageDriver interface.
+func (d *StorageDriver) PluginTypeID() string { return "multi" }
+
+// Init implements the keppel.StorageDriver interface.
+func (d *StorageDriver) Init(ad keppel.AuthDriver, cfg keppel.Configuration) error {
```
majewsky: Should check if `d.Phase` contains one of the valid values.

internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,415 @@
...
+				err := d.newDriver.AppendToBlob(ctx, account, storageID, 1, Some(sizeBytes), newDriverPR)
+				newDriverPR.CloseWithError(err)
+				if err != nil {
+					logg.Error("multi-driver: while copying blob %s to new driver: %s", storageID, err.Error())
+					return
+				}
+				err = d.newDriver.FinalizeBlob(ctx, account, storageID, 1)
+				if err != nil {
+					logg.Error("multi-driver: while finalizing blob %s in new driver: %s", storageID, err.Error())
+				}
```
majewsky: This will blow up if `sizeBytes` is larger than the maximum acceptable segmentSize in `newDriver`. We need to reuse the chunking logic from `Processor.AppendToBlob()` here, which may necessitate moving it into `package keppel` to avoid an import cycle.

majewsky (changes requested):

internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,415 @@
...
```
majewsky: Error branches should call `d.newDriver.AbortBlobUpload()` to reduce the risk of leaving behind debris after an error.

internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,415 @@
...
+	default:
+		panic(fmt.Sprintf("multi-driver: unexpected phase %q", d.Phase))
+	}
+}
+
+// CanSetupAccount implements the keppel.StorageDriver interface.
+func (d *StorageDriver) CanSetupAccount(ctx context.Context, account models.ReducedAccount) error {
+	switch d.Phase {
+	case CopyPhase:
+		return d.oldDriver.CanSetupAccount(ctx, account)
```
majewsky: Should check both sides, in order to forbid accounts that would be illegal after the migration.

majewsky (changes requested):

internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,465 @@
...
+// In CopyPhase it appends to both storages in parallel.
+func (d *StorageDriver) AppendToBlob(ctx context.Context, account models.ReducedAccount, storageID string, chunkNumber uint32, chunkLength Option[uint64], chunk io.Reader) error {
+	switch d.Phase {
+	case CopyPhase:
+		pr, pw := io.Pipe()
+		tee := io.TeeReader(chunk, pw)
+
+		errCh := make(chan error, 1)
+
+		go func() {
```
majewsky, line 115: Should `defer close(errCh)` for extra safety in case the goroutine does not get to the end.

internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,465 @@
...
+
+const chunkSizeBytes = 500 << 20 // 500 MiB
+
+// appendToBlob contains the logic for splitting `contents` (containing `lengthBytes`) into chunks of `chunkSizeBytes` max.
+// NOTE: This function is written such that `action` is called at least once, even when `contents` is empty.
+func appendToBlob(ctx context.Context, sd keppel.StorageDriver, account models.ReducedAccount, upload *models.Upload, contents io.Reader, lengthBytes uint64) error {
+	action := func(chunk io.Reader, chunkLengthBytes uint64) error {
+		upload.NumChunks++
+		upload.SizeBytes += chunkLengthBytes
+		return sd.AppendToBlob(ctx, account, upload.StorageID, upload.NumChunks, Some(chunkLengthBytes), chunk)
```
majewsky: The only real effect of `upload` is to count up `upload.NumChunks()` as an argument for `sd.AppendToBlob`. This can be replaced with a simple counter variable.

~~Also, `action` can be inlined. It is not inlined in the original implementation because two different branches use it, but here only one branch remains.~~

internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,465 @@
...
+func (d *StorageDriver) ReadBlob(ctx context.Context, account models.ReducedAccount, storageID string) (io.ReadCloser, uint64, error) {
+	switch d.Phase {
+	case CopyPhase:
+		return d.oldDriver.ReadBlob(ctx, account, storageID)
+	default:
+		panic(fmt.Sprintf("multi-driver: unexpected phase %q", d.Phase))
+	}
+}
+
+const chunkSizeBytes = 500 << 20 // 500 MiB
```
majewsky: This constant should be placed in `package keppel` and shared with its other instance.

```suggestion
// ChunkSizeBytes is the maximum size for chunks that Keppel writes into a [StorageDriver].
const ChunkSizeBytes = 500 << 20 // 500 MiB
```

internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,465 @@
...
+
+			go func() {
+				upload := models.Upload{
+					StorageID: keppel.GenerateStorageID(),
+					SizeBytes: 0,
+					NumChunks: 0,
+				}
+
+				err = appendToBlob(ctx, d.newDriver, account, &upload, newDriverPR, sizeBytes)
+				newDriverPR.CloseWithError(err)
```
majewsky: I understand the point of this setup: to stream from old->new while simulatenously returning the read contents to the caller in a nonblocking way. However, since we are explicitly in the "for validation" callpath here, we can afford to take our time to stream old->new and then return a fresh reader for the same payload on the new side, thus avoiding the pipe/copy/goroutine complexity.

majewsky (changes requested):

internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,445 @@
...
+// In CopyPhase it tries the new driver first. If the blob does not exist there,
+// it reads from the old driver and simultaneously copies the data to the new driver
+// so the blob is migrated on first validation.
+func (d *StorageDriver) ReadBlobForValidation(ctx context.Context, account models.ReducedAccount, storageID string) (io.ReadCloser, uint64, error) {
+	switch d.Phase {
+	case CopyPhase:
+		reader, sizeBytes, err := d.newDriver.ReadBlob(ctx, account, storageID)
+		if errors.Is(err, os.ErrNotExist) {
+			reader, sizeBytes, err = d.oldDriver.ReadBlob(ctx, account, storageID)
+			if err != nil {
```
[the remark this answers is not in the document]
SuperSandro2000: @majewsky that should not be a problem for us, right? Or do we want to write tests e2e tests for this?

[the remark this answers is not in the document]
majewsky: See my note on the backlog ticket for this:

> Existing storage drivers shall be amended to wrap respective backend-specific errors in this type as necessary (e.g. in Swift, errors with `schwift.Is(err, http.StatusNotFound)`; or in the filesystem driver, errors with `os.IsNotExist(err)`).

internal/keppel/storage_driver.go
```diff
@@ -161,3 +166,21 @@ func GenerateStorageID() string {
 	}
 	return hex.EncodeToString(buf)
 }
+
+// NotFoundInStorageError is a generic error that the backend could not find the specific entry
```
majewsky:
```suggestion
// NotFoundInStorageError is a generic error indicating that a StorageDriver could not find the requested entry.
// The actual backend error is wrapped in this type.
```

internal/keppel/storage_driver.go
```diff
@@ -161,3 +166,21 @@ func GenerateStorageID() string {
...
+type NotFoundInStorageError struct {
+	Inner error
+}
+
+func (e NotFoundInStorageError) Error() string {
```
majewsky, line 177:
```suggestion
// Error implements the builtin/error interface.
func (e NotFoundInStorageError) Error() string {
```

internal/keppel/storage_driver.go
```diff
@@ -161,3 +166,21 @@ func GenerateStorageID() string {
...
+	return e.Inner.Error()
+}
+
+func (e NotFoundInStorageError) Is(target error) bool {
```
majewsky, line 182:
```suggestion
// Is implements the interface implied by the documentation for [errors.Is].
func (e NotFoundInStorageError) Is(target error) bool {
```

internal/keppel/storage_driver.go
```diff
@@ -161,3 +166,21 @@ func GenerateStorageID() string {
...
+	_, ok := target.(NotFoundInStorageError)
+	return ok
+}
+
+func (e NotFoundInStorageError) Unwrap() error {
```
majewsky, line 188:
```suggestion
// Unwrap implements the interface implied by the documentation for package errors.
func (e NotFoundInStorageError) Unwrap() error {
```

internal/drivers/filesystem/storage.go
```diff
@@ -88,31 +91,50 @@ func (d *StorageDriver) AppendToBlob(ctx context.Context, account models.Reduced
...
+	if os.IsNotExist(err) {
+		return nil, 0, keppel.NotFoundInStorageError{Inner: err}
+	}
 	if err != nil {
 		return nil, 0, err
 	}
 	stat, err := f.Stat()
+	if os.IsNotExist(err) {
+		return nil, 0, keppel.NotFoundInStorageError{Inner: err}
+	}
```
majewsky, line 123: This error path does not `f.Close()`. To avoid code duplication:
```suggestion
	if os.IsNotExist(err) {
		err = keppel.NotFoundInStorageError{Inner: err}
	}
```

internal/drivers/filesystem/storage.go
```diff
@@ -121,13 +143,26 @@ func (d *StorageDriver) URLForBlob(ctx context.Context, account models.ReducedAc
 // DeleteBlob implements the keppel.StorageDriver interface.
 func (d *StorageDriver) DeleteBlob(ctx context.Context, account models.ReducedAccount, storageID string) error {
 	path := d.getBlobPath(account, storageID)
-	return os.Remove(path)
+	err := os.Remove(path)
+	if os.IsNotExist(err) {
+		return keppel.NotFoundInStorageError{Inner: err}
+	}
+	return err
```
majewsky: This particular sequence is repeated a lot, so I would like to factor it out, esp. to make it easier to add additional custom error types for other situations later.
```suggestion
	return classifyError(err)
```

Where

```go
func classifyError(err error) error {
	if os.IsNotExist(err) {
		return keppel.NotFoundInStorageError{Inner: err}
	} else {
		return err
	}
}
```

Same for all other instances in this file.

internal/drivers/openstack/swift.go
```diff
@@ -250,13 +254,19 @@ func (d *swiftDriver) FinalizeBlob(ctx context.Context, account models.ReducedAc
 		},
 		&schwift.TruncateOptions{DeleteSegments: false},
 	)
+	if schwift.Is(err, http.StatusNotFound) {
+		return keppel.NotFoundInStorageError{Inner: err}
+	}
 	if err != nil {
 		return err
 	}
```
majewsky: Same remark as for the `filesystem` driver. This should be
```suggestion
	if err != nil {
		return classifyError(err)
	}
```
with a suitable `classifyError` implementation, and same for all other instances where the "is not found" check is added in this file.

internal/drivers/openstack/swift.go
```diff
@@ -308,13 +321,29 @@ func (d *swiftDriver) ReadBlob(ctx context.Context, account models.ReducedAccoun
...
 	return reader, hdr.SizeBytes().Get(), err
 }
 
+// ReadBlobForValidation implements the keppel.StorageDriver interface.
+func (d *swiftDriver) ReadBlobForValidation(ctx context.Context, account models.ReducedAccount, storageID string) (io.ReadCloser, uint64, error) {
+	reader, size, err := d.ReadBlob(ctx, account, storageID)
+	if schwift.Is(err, http.StatusNotFound) {
+		return nil, 0, keppel.NotFoundInStorageError{Inner: err}
+	}
+	return reader, size, err
```
majewsky: ReadBlob already wraps the error if necessary:
```suggestion
	return d.ReadBlob(ctx, account, storageID)
```
Same for ReadManifestForValidation.

internal/drivers/trivial/storage.go
```diff
@@ -205,7 +205,7 @@ func (d *StorageDriver) FinalizeBlob(ctx context.Context, account models.Reduced
 	defer d.blobsMutex.RUnlock()
 	_, exists := d.blobs[k]
 	if !exists {
-		return errNoSuchBlob
+		return keppel.NotFoundInStorageError{Inner: errNoSuchBlob}
```
majewsky, line 208: This repeated change can be simplified by wrapping at the definition site of `errNoSuchBlob` et al, i.e.

```go
var errNoSuchBlob = errors.New("no such blob")
```

can be rewritten into

```go
var errNoSuchBlob = keppel.NotFoundInStorageError{Inner: errors.New("no such blob")}
```

Same for the other `errNoSuch...` variables.

majewsky (changes requested):
Please split the introduction of `keppel.NotFoundInStorageError` into a separate PR. Here are comments specifically on that part:

SuperSandro2000:
The new error type and the feedback all went into #761

internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,654 @@
...
+
+	var numChunks uint32
+	err = appendToBlob(ctx, d.newDriver, account, storageID, reader, sizeBytes, &numChunks)
+	if err != nil {
+		err = fmt.Errorf("while copying blob %s to new driver: %w", storageID, err)
+		err2 := d.newDriver.AbortBlobUpload(ctx, account, storageID, numChunks)
+		if err2 != nil {
+			return errors.Join(err, fmt.Errorf("while aborting copying blob %s to new driver an additional error has occurred: %w", storageID, err2))
+		}
+		return err
```
majewsky: errors.Join() is annoying because it fuses errors with newlines, which messes up our logging format (our  OpenSearch instances break up logs that span across multiple lines into separate records).

I have a new API in the works in gg that will be useful here, so flagging this for myself as a reminder to cut a new gg release.

internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,654 @@
...
+			if migrateErr == nil {
+				reader, sizeBytes, err = d.newDriver.ReadBlob(ctx, account, storageID)
+				if err != nil {
+					return nil, 0, fmt.Errorf("while reading from new driver: %w", err)
+				}
+			} else {
+				reader, sizeBytes, err = d.oldDriver.ReadBlob(ctx, account, storageID)
+				if err != nil {
+					return nil, 0, fmt.Errorf("while reading from old driver: %w", err)
+				}
```
majewsky: Should call ReadBlobForValidation, for consistency with the branches for the other phases.

internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,654 @@
...
+	case copyPhase:
+		oldErr := d.oldDriver.DeleteBlob(ctx, account, storageID)
+		oldNotFound := errors.Is(oldErr, keppel.NotFoundInStorageError{})
+
+		newErr := d.newDriver.DeleteBlob(ctx, account, storageID)
+		newNotFound := errors.Is(newErr, keppel.NotFoundInStorageError{})
+
+		if oldNotFound && newNotFound {
+			return keppel.NotFoundInStorageError{}
+		}
```
majewsky, line 316: `keppel.NotFoundInStorageError{}` on its own does not have an error message, so we should forward the underlying errors:
```suggestion
		if oldNotFound && newNotFound {
			return fmt.Errorf("missing in both drivers: old driver: %w, new driver: %w", oldErr, newErr)
		}
```

internal/drivers/multi/storage.go
majewsky, line 1: There should be documentation for this driver. You can leave this to me if you're not into technical writing; I'm just putting this here so we don't forget.

internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,654 @@
...
+	case cleanupPhase:
+		err := d.oldDriver.DeleteTrivyReport(ctx, account, repoName, manifestDigest, format)
+		if err != nil && !errors.Is(err, keppel.NotFoundInStorageError{}) {
+			return nil, err
+		}
+		if err == nil {
+			cleanedUpCounter.With(prometheus.Labels{"type": "trivy_reports"}).Inc()
+		}
+
+		return d.newDriver.ReadTrivyReport(ctx, account, repoName, manifestDigest, format)
```
majewsky: We don't need a delete-on-old here. Trivy reports are regularly rewritten, so the delete during `cleanupPhase` in WriteTrivyReport serves the same purpose that the ReadFooForValidation calls serve for blobs and manifests.

internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,654 @@
...
+	case copyPhase:
+		oldErr := d.oldDriver.DeleteTrivyReport(ctx, account, repoName, manifestDigest, format)
+		oldNotFound := errors.Is(oldErr, keppel.NotFoundInStorageError{})
+
+		newErr := d.newDriver.DeleteTrivyReport(ctx, account, repoName, manifestDigest, format)
+		newNotFound := errors.Is(newErr, keppel.NotFoundInStorageError{})
+
+		if oldNotFound && newNotFound {
+			return keppel.NotFoundInStorageError{}
+		}
```
majewsky, line 520: Same note as above: This should have an error message combining `oldErr` and `newErr`.

internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,654 @@
...
+		objectsGauge.With(prometheus.Labels{"account": string(account.Name), "type": "trivy_reports"}).Set(float64(len(trivyReports)))
+
+		return d.newDriver.ListStorageContents(ctx, account)
+	case finalizePhase:
+		_, _, _, err := d.oldDriver.ListStorageContents(ctx, account)
+		if !errors.Is(err, keppel.NotFoundInStorageError{}) {
+			return nil, nil, nil, fmt.Errorf("old driver contains unexpected account %s", account.Name)
+		}
+
+		return d.newDriver.ListStorageContents(ctx, account)
```
majewsky, line 576: The spec calls for ListStorageContents to call CleanupAccount on the old size in the "finalize" phase. Failing on a not-found in ListStorageContents on old does not make sense here, since that is the intended target state: We _want_ for all accounts to be missing on the old side eventually. The correct behavior is to try CleanupAccount, and only propagate errors if they are not of the NotFoundInStorage variety.

majewsky (changes requested):
These are my notes for the file `internal/drivers/multi/storage.go`. Overall, I'm very happy with the overall implementation quality; most of my notes are just papercuts and nitpicks. Once these are addressed and the PR is rebased on main, I will have a look at the remaining bits and pieces, but we should be good to go here soon.

internal/drivers/multi/storage.go
majewsky: Update: `go get -u go.xyrillian.de/gg@v1.12.0` and then
```suggestion
		err2 := d.newDriver.AbortBlobUpload(ctx, account, storageID, numChunks)
		return errext.WithCleanup(err, "newDriver.AbortBlobUpload", err2)
```
Ref: <https://pkg.go.dev/go.xyrillian.de/gg/errext#WithCleanup>

internal/drivers/multi/storage.go
```diff
@@ -0,0 +1,638 @@
...
+// PluginTypeID implements the keppel.StorageDriver interface.
+func (d *StorageDriver) PluginTypeID() string { return "multi" }
+
+// Init implements the keppel.StorageDriver interface.
+func (d *StorageDriver) Init(ctx context.Context, ad keppel.AuthDriver, cfg keppel.Configuration) error {
+	if d.Phase != copyPhase && d.Phase != cleanupPhase && d.Phase != finalizePhase {
+		return fmt.Errorf("phase contains invalid name %s, only copy, cleanup or finalize are allowed", d.Phase)
+	}
+
+	logg.Debug("initializing multi storage driver old driver: %q", d.OldParams)
```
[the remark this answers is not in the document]
majewsky: This is a valid concern. Let's get rid of these log lines; they don't add anything.

majewsky (approved):

PR #753, opened 2026-07-20
api/registry: reflect firstpulls from replica to external replica

majewsky:
When pulling from a replica account whose upstream is an external replica account, users may now initiate replication on the replica account, and Keppel will propagate it to the external replica in one go. Users thus will no longer have to manually interact with the external replica account first. Pulling from any replica will have the desired effect.

Most of the work here is from propagating the UserType of the original requester to the Keppel holding the external replica account, so that we can properly tell anonymous users that they need to be authenticated to proceed.

### DIFF
--- a/internal/api/registry/api.go
+++ b/internal/api/registry/api.go
@@ -267,12 +267,36 @@ func (a *API) checkAccountAccess(w http.ResponseWriter, r *http.Request, strateg
 		return nil, nil, nil, nil
 	}
 
+	requestingUserType := authz.UserIdentity.UserType()
+	if authz.UserIdentity.UserType() == keppel.PeerUser {
+		// if another Keppel is performing requests on a user's behalf,
+		// access checks are usually contingent on that user's type
+		// (cf. long comment in mayReplicateManifest())
+		if userType, ok := keppel.ParseUserType(r.Header.Get("X-Keppel-Requesting-User-Type")).Unpack(); ok {
+			requestingUserType = userType
+		}
+	}
+
 	canCreateRepoIfMissing := false
 	switch strategy {
 	case createRepoIfMissing:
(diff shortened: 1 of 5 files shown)


internal/api/registry/api.go
```diff
@@ -267,12 +267,34 @@ func (a *API) checkAccountAccess(w http.ResponseWriter, r *http.Request, strateg
 		return nil, nil, nil, nil
 	}
 
+	requestingUserType := authz.UserIdentity.UserType()
+	if authz.UserIdentity.UserType() == keppel.PeerUser {
+		// if another Keppel is performing requests on a user's behalf,
+		// access checks are usually contingent on that user's type
+		// (cf. long comment in mayReplicateManifest())
+		requestingUserType = keppel.UserType(r.Header.Get("X-Keppel-Requesting-User-Type"))
+	}
```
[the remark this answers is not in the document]
majewsky: Added explicit serializing and parsing.

internal/processor/manifests.go
```diff
@@ -824,6 +832,7 @@ func (p *Processor) downloadManifestViaInboundCache(ctx context.Context, account
 	// cache miss -> download from actual upstream registry
 	manifestBytes, manifestMediaType, err = c.DownloadManifest(ctx, ref, &client.DownloadManifestOpts{
 		DoNotCountTowardsLastPulled: true,
+		ExtraHeaders:                extraHeaders,
```
SuperSandro2000, line 835: This can be removed again

majewsky: No, this is relevant.

SuperSandro2000 (dismissed):

SuperSandro2000 (approved):

PR #765, opened 2026-08-04
fix container creation when using Ceph

majewsky:
In the `UseServiceUserProject = true` path, we never get to the part where we do Create() on the container to set tempurl keys, so the containers are never actually created.

### DIFF
--- a/internal/drivers/openstack/swift.go
+++ b/internal/drivers/openstack/swift.go
@@ -148,6 +148,12 @@ func (d *swiftDriver) getBackendConnection(ctx context.Context, account models.R
 
 	// if all containers are in the same account, use that account's tempurl key
 	if d.UseServiceUserProject {
+		// we do not have the Create() below that writes the container tempurl key,
+		// so we need to explicitly make sure that the backing container exists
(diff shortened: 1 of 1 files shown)


internal/drivers/openstack/swift.go
```diff
@@ -148,7 +148,10 @@ func (d *swiftDriver) getBackendConnection(ctx context.Context, account models.R
 
 	// if all containers are in the same account, use that account's tempurl key
 	if d.UseServiceUserProject {
-		return c, d.mainTempURLKey, nil
+		// we do not have the Create() below that writes the container tempurl key,
+		// so we need to explicitly make sure that the backing container exists
+		_, err := c.EnsureExists(ctx)
+		return c, d.mainTempURLKey, err
```
SuperSandro2000:
```suggestion
		_, err := c.EnsureExists(ctx)
		if err != nil {
		  return nil, nil, err
		}
		return c, d.mainTempURLKey, nil
```

SuperSandro2000 (approved):

PR #772, opened 2026-08-10
Add self healing to ValidateExistingBlob and validateManifest

SuperSandro2000:


internal/api/registry/replication_test.go
```diff
@@ -598,3 +601,71 @@ func TestReplicationSuccessfulFromNVCRio(t *testing.T) {
...
+			if err != nil {
+				t.Fatal("blob not readable from storage after self-healing:", err)
+			}
+			readCloser.Close()
+			if sizeBytes != reloadedBlob.SizeBytes {
+				t.Errorf("expected blob size %d, but got %d", reloadedBlob.SizeBytes, sizeBytes)
+			}
+
+			s1.Clock.StepBy(time.Second)
+			expectBlobExists(t, s2, tokenHeaders, "test1/foo", image.Layers[0])
```
majewsky, line 658: Just calling `expectBlobExists` is enough here, that already reads the blob in full.

internal/processor/blobs.go
```diff
@@ -34,6 +34,32 @@ func (p *Processor) ValidateExistingBlob(ctx context.Context, account models.Red
 	}
 
 	readCloser, _, err := p.sd.ReadBlobForValidation(ctx, account, blob.StorageID)
+	// If we cannot find the blob and the account is a replication from somewhere else try to get it from there
+	if errors.Is(err, keppel.NotFoundInStorageError{}) && (account.ExternalPeerURL != "" || account.UpstreamPeerHostName != "") {
+		// find any repo that mounts this blob (needed by ReplicateBlob to identify the upstream repo)
+		repo, repoErr := models.ReducedRepositoryStore.SelectOne(ctx, p.db,
+			`SELECT r.id, r.account_name, r.name FROM repos r JOIN blob_mounts bm ON r.id = bm.repo_id WHERE bm.blob_id = $1 LIMIT 1`,
+			blob.ID,
+		)
```
majewsky: You can avoid having to list out the target columns with SelectOneWhere by rewriting the join into a subquery:
```suggestion
		repo, repoErr := models.ReducedRepositoryStore.SelectOneWhere(ctx, p.db,
			`id = (SELECT repo_id FROM blob_mounts WHERE blob_id = $1)`,
			blob.ID,
		)
```

internal/processor/manifests.go
```diff
@@ -177,8 +177,15 @@ func (p *Processor) ValidateAndStoreManifest(ctx context.Context, account models
...
 // ValidateExistingManifest validates the given manifest that already exists in the DB.
-func (p *Processor) ValidateExistingManifest(ctx context.Context, account models.ReducedAccount, repo models.ReducedRepository, manifest *models.Manifest) error {
+func (p *Processor) ValidateExistingManifest(ctx context.Context, account models.ReducedAccount, repo models.ReducedRepository, manifest *models.Manifest, tagPolicies []keppel.TagPolicy, actx keppel.AuditContext) error {
 	manifestBytes, err := p.sd.ReadManifestForValidation(ctx, account, repo.Name, manifest.Digest)
+	// If we cannot find the manifest and the account is a replication from somewhere else try to get it from there
+	if errors.Is(err, keppel.NotFoundInStorageError{}) && (account.ExternalPeerURL != "" || account.UpstreamPeerHostName != "") {
+		manifest, manifestBytes, err = p.ReplicateManifest(ctx, account, repo, models.ManifestReference{Digest: manifest.Digest}, tagPolicies, actx)
+		if err != nil {
+			return fmt.Errorf("manifest could not be found while validating and replication failed: %w", err)
+		}
```
majewsky: We should not swallow the original error:
```suggestion
		var replicationErr error
		manifest, manifestBytes, replicationErr = p.ReplicateManifest(ctx, account, repo, models.ManifestReference{Digest: manifest.Digest}, tagPolicies, actx)
		if replicationErr != nil {
			return fmt.Errorf("%w (additional error while trying to self-heal by replicating this manifest: %w)", err, replicationErr)
		}
```

majewsky (changes requested):
Three small things:

majewsky (approved):

PR #773, opened 2026-08-11
Cleanup pending_blobs after OOMkill by emitting a heartbeat every minute

SuperSandro2000:
Alternative to #768 which requires less active database connections

Closes #768

### DIFF
--- a/internal/models/pending_blob.go
+++ b/internal/models/pending_blob.go
@@ -16,6 +16,7 @@ type PendingBlob struct {
 	Digest       digest.Digest `db:"digest"`
 	Reason       PendingReason `db:"reason"`
 	PendingSince time.Time     `db:"since"`
+	LastSeenAt   time.Time     `db:"last_seen_at"`
 }
 
 // PendingBlobStore provides loading and storing of [PendingBlob] objects from the DB.
--- a/internal/processor/blobs.go
+++ b/internal/processor/blobs.go
@@ -124,6 +124,7 @@ func (p *Processor) ReplicateBlob(ctx context.Context, blob models.Blob, account
 		Digest:       blob.Digest,
 		Reason:       models.PendingBecauseOfReplication,
 		PendingSince: p.timeNow(),
+		LastSeenAt:   p.timeNow(),
 	}
 	err := models.PendingBlobStore.Insert(ctx, p.db, &pendingBlob)
 	if err != nil {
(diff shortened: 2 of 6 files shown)


internal/tasks/pending_blobs.go
```diff
@@ -0,0 +1,46 @@
...
+			maxHeartbeatAt := j.timeNow().Add(-3 * time.Minute)
+			return models.PendingBlobStore.SelectOne(ctx, j.db, stalePendingBlobsQuery, maxHeartbeatAt)
+		},
+		ProcessTask: j.sweepPendingBlobs,
+	}).Setup(registerer)
+}
+
+func (j *Janitor) sweepPendingBlobs(ctx context.Context, pendingBlob models.PendingBlob, _ prometheus.Labels) error {
+	logg.Info("cleaning up stale pending blob %s in account %s", pendingBlob.Digest, pendingBlob.AccountName)
+	return models.PendingBlobStore.Delete(ctx, j.db, pendingBlob)
```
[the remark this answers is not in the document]
SuperSandro2000: If the heartbeat did not happen for 3 minutes, it is definitely dead.

internal/models/pending_blob.go
```diff
@@ -16,6 +16,7 @@ type PendingBlob struct {
 	Digest       digest.Digest `db:"digest"`
 	Reason       PendingReason `db:"reason"`
 	PendingSince time.Time     `db:"since"`
+	HeartbeatAt  time.Time     `db:"heartbeat_at"`
```
majewsky: While "heartbeat" is the correct technical term for this sort of thing, usually a heartbeat field will have a more specific name. Besides, "heartbeat_at" can be read as "the next heartbeat will be at X", which is not the intent. Our "somethingSomethingAt" columns usually have "last" or "next" in them to clarify this, so I suggest:
```suggestion
	LastSeenAt   time.Time     `db:"last_seen_at"`
```

internal/processor/blobs.go
```diff
@@ -151,12 +152,40 @@ func (p *Processor) ReplicateBlob(ctx context.Context, blob models.Blob, account
 		}
 	}()
 
+	// start a heartbeat so that the janitor does not clean up this entry while it is still replicating
+	replicationCtx, cancelReplication := context.WithCancel(ctx)
+	defer cancelReplication()
+	go func() {
+		ticker := time.NewTicker(1 * time.Minute)
```
majewsky: We can go with a tighter interval here. 15 seconds should still not cause too much dead row spam, and it improves the user experience if an aborted replication can be retried faster:
```suggestion
		ticker := time.NewTicker(15 * time.Second)
```

internal/tasks/pending_blobs.go
```diff
@@ -0,0 +1,46 @@
...
+	return (&jobloop.ProducerConsumerJob[models.PendingBlob]{
+		Metadata: jobloop.JobMetadata{
+			ReadableName: "cleanup pending blobs",
+			CounterOpts: prometheus.CounterOpts{
+				Name: "keppel_cleanup_pending_blobs",
+				Help: "Counter for garbage collections on orphaned pending blobs.",
+			},
+		},
+		DiscoverTask: func(ctx context.Context, _ prometheus.Labels) (models.PendingBlob, error) {
+			maxHeartbeatAt := j.timeNow().Add(-3 * time.Minute)
```
majewsky: Concordantly:
```suggestion
			maxLastSeenAt := j.timeNow().Add(-1 * time.Minute)
```

internal/tasks/pending_blobs.go
```diff
@@ -0,0 +1,46 @@
...
+var stalePendingBlobsQuery = sqlext.SimplifyWhitespace(`
+	SELECT * FROM pending_blobs
+	WHERE heartbeat_at < $1
+	ORDER BY heartbeat_at ASC
+	LIMIT 1
+`)
+
+// CleanupPendingBlobsJob is a jobloop.Job. Each task finds one pending_blob whose heartbeat is older than 3 minutes and deletes it.
+func (j *Janitor) CleanupPendingBlobsJob(registerer prometheus.Registerer) jobloop.Job { //nolint:dupl // false positive
+	return (&jobloop.ProducerConsumerJob[models.PendingBlob]{
```
majewsky: This does not need to be a ProducerConsumerJob. You can turn this into a CronJob and do everything in one query:

```sql
DELETE FROM pending_blobs WHERE last_seen_at < $1 RETURNING *;
```

The RETURNING clause allows you to `.Select()` on this query and get the deleted records for the purpose of logging.

majewsky (changes requested):
The heartbeat loop itself looks good. A couple of notes regarding all the other pieces:

majewsky (approved):

PR #774, opened 2026-08-13
Mention which image was rejected because the docker client cannot

SuperSandro2000:


internal/api/registry/manifests.go
```diff
@@ -251,8 +251,8 @@ func mayReplicateManifest(account models.ReducedAccount, authz auth.Authorizatio
 
 	// when replicating from external, only authenticated users can trigger the replication
 	if account.ExternalPeerURL != "" && userType != keppel.RegularUser {
-		if !authz.ScopeSet.AllowsAnonymousFirstPullOn(mux.Vars(r)["repository"]) {
-			rerr := keppel.ErrDenied.With("image does not exist here, and anonymous users may not replicate images")
+		if !authz.ScopeSet.AllowsAnonymousFirstPullOn(repo.FullName()) {
+			rerr := keppel.ErrDenied.With(fmt.Sprintf("image %q does not exist here, and anonymous users may not replicate images", repo.FullName()+":"+reference.String()))
```
majewsky: This will produce the wrong syntax (`repo:digest` instead of `repo@digest`) when pulling by digest instead of by tag:
```suggestion
			rerr := keppel.ErrDenied.With(fmt.Sprintf("image %q does not exist here, and anonymous users may not replicate images", models.ImageReference{RepoName:repo.FullName(), Reference:reference}))
```
This suggestion is also somewhat wrong because, if `ImageReference.Host == ""` like in the instance we build, stringification will strip a leading `library/` from the repo name under the assumption that we're referring to an image on Docker Hub. I find that acceptable in this case.

majewsky (changes requested):

internal/api/registry/manifests.go
SuperSandro2000: We have the Host from the request, so we can throw that into the mix.

majewsky (approved):

PR #780, opened 2026-08-18
Allow "anonymous_first_pull" RBACs on replica accounts of external replicas

SuperSandro2000:


internal/api/registry/replication_test.go
```diff
@@ -514,6 +514,77 @@ func TestReplicationChainedFromExternalToInternalReplica(t *testing.T) {
...
+			test.MustExec(t, s1.DB, `UPDATE accounts SET external_peer_url = $2 WHERE name = $1`,
+				"test1", "registry-external.example.org")
+
+			// enable anonymous first pull on the external replica
+			test.MustExec(t, s1.DB, `UPDATE accounts SET rbac_policies_json = $2 WHERE name = $1`, "test1",
+				test.ToJSON([]keppel.RBACPolicy{{
+					RepositoryPattern: ".*",
+					Permissions:       []keppel.RBACPermission{keppel.RBACAnonymousPullPermission, keppel.RBACAnonymousFirstPullPermission},
+				}}),
+			)
```
majewsky: Below, we're configuring rbac_policies using the API for `s2`. We should do the same for `s1` a) for consistency and b) for additional test coverage of the updated RBAC policy validation rules.

internal/api/registry/replication_test.go
```diff
@@ -514,6 +514,77 @@ func TestReplicationChainedFromExternalToInternalReplica(t *testing.T) {
...
+							"match_repository": ".*",
+							"permissions":      []string{"anonymous_pull", "anonymous_first_pull"},
+						}},
+					},
+				}),
+			).ExpectStatus(t, http.StatusOK)
+
+			// an anonymous user pulling from s2 triggers replication both into s1 and s2
+			anonTokenHeaders := s2.GetAnonTokenHeaders(t, "repository:test1/foo", []string{"pull", "anonymous_first_pull"})
+			expectManifestExists(t, s2, anonTokenHeaders, "test1/foo", image.Manifest, "latest")
```
majewsky, line 591: Since the comment calls out that we replicate "both into s1 and s2", we should have an additional `expectManifestExists()` check for s1.

docs/api-spec.md
```diff
@@ -178,7 +178,7 @@ The following fields may be returned:
...
-| `accounts[].rbac_policies[].permissions` | list of strings | The permissions granted by the RBAC policy. Acceptable values include `pull`, `push`, `delete`, `anonymous_pull` and `anonymous_first_pull`. When `pull`, `push` or `delete` are included, `match_username` is not empty. When `anonymous_pull` or `anonymous_first_pull` is included, `match_username` is empty. `anonymous_first_pull` is only relevant for external replica accounts and allows unauthenticated users to replicate tags. It should always be combined with an appropriate `match_*` rule. |
+| `accounts[].rbac_policies[].permissions` | list of strings | The permissions granted by the RBAC policy. Acceptable values include `pull`, `push`, `delete`, `anonymous_pull` and `anonymous_first_pull`. When `pull`, `push` or `delete` are included, `match_username` is not empty. When `anonymous_pull` or `anonymous_first_pull` is included, `match_username` is empty. `anonymous_first_pull` is only relevant for replica accounts and allows unauthenticated users to replicate tags. It should always be combined with an appropriate `match_*` rule. |
```
majewsky: Not a comment for the change you made, but I noticed a phrasing error on this line that we can fix while we're at it:

```suggestion
| `accounts[].rbac_policies[].permissions` | list of strings | The permissions granted by the RBAC policy. Acceptable values include `pull`, `push`, `delete`, `anonymous_pull` and `anonymous_first_pull`. When `pull`, `push` or `delete` are included, `match_username` is not empty. When `anonymous_pull` or `anonymous_first_pull` is included, `match_username` is empty. `anonymous_first_pull` is only relevant for replica accounts and allows unauthenticated users to replicate manifests and tags. It should always be combined with an appropriate `match_*` rule. |
```

majewsky (changes requested):

internal/api/registry/replication_test.go
SuperSandro2000: We then also need to add test.WithKeppelAPI to s1.

majewsky (approved):

PR #789, opened 2026-08-28
add accounts.anon_rbac_policies_json

majewsky:
As the comments hint at, the final intent for this is to skip issuing signed tokens for anonymous users. Instead, in the common case, anonymous users should receive a token that is basically just equal to `base64(join(audience, repo_name))`. There is precedent for this with GHCR, which also issues simple Base64-encoded strings only for anonymous users, so this ought not break existing users:

```
$ curl -s 'https://ghcr.io/token?service=ghcr.io&scope=repository:home-assistant/home-assistant:pull' | tee /tmp/resp.json
{"token":"djE6aG9tZS1hc3Npc3RhbnQvaG9tZS1hc3Npc3RhbnQ6MTc4NzkwODAxNTc5OTYxMjExMA=="}
$ </tmp/resp.json jq -r .token | base64 -d
v1:home-assistant/home-assistant:1787908015799612110
```

This would allow us to skip verifying the cryptographic signature on a JWT-style token, which currently makes up a significant amount of the CPU cost for common hot paths like GetBlob and GetManifest. However, we then need to recheck AuthZ during these hot paths. This will be made possible by including the relevant RBAC policies for anonymous users in the ReducedAccount.

I have made an effort to make payloads in the new field as compact as possible, and beyond that, there is a protection to keep the field from growing too large even if a user configures a large amount of RBAC policies: The worst that will happen is that the field is not populated at all, and AuthZ will instead issue a normal JWT-style token to the anonymous user.

There is a problem here, in that we don't have an easy way to populate this field for existing accounts. Since we currently still have direct DB access to all prod deployments, I plan to address this migration problem with a script that GETs and PUTs each account once, thus triggering the relevant codepath that fills the new column.


internal/keppel/rbac_policy.go
```diff
@@ -152,3 +171,78 @@ func ParseRBACPoliciesField(buf []byte) ([]RBACPolicy, error) {
...
+		perms = append(perms, "!f")
+	}
+	return json.MarshalEncode(enc, serializedAnonymousRBACPolicy{
+		CidrPattern:       a.cidrPattern,
+		RepositoryPattern: a.repositoryPattern,
+		Permissions:       strings.Join(perms, ","),
+	})
+}
+
+// UnmarshalJSONFrom implements the [json.UnmarshalerForm] interface.
```
SuperSandro2000, line 221:
```suggestion
// UnmarshalJSONFrom implements the [json.UnmarshalerFrom] interface.
```

internal/processor/accounts.go
```diff
@@ -158,19 +158,32 @@ func (p *Processor) CreateOrUpdateAccount(ctx context.Context, account keppel.Ac
...
+		anonPolicies := []keppel.AnonymousRBACPolicy{}
 		for idx, policy := range account.RBACPolicies {
-			err := policy.ValidateAndNormalize(replicationStrategy)
+			anonPolicy, err := policy.ValidateAndNormalize(replicationStrategy)
 			if err != nil {
 				return models.Account{}, keppel.AsRegistryV2Error(err).WithStatus(http.StatusUnprocessableEntity)
 			}
 			account.RBACPolicies[idx] = policy
+
+			if p, ok := anonPolicy.Unpack(); ok {
```
SuperSandro2000, line 174:
```suggestion
			if pol, ok := anonPolicy.Unpack(); ok {
```
to not shadow *Processor

internal/keppel/rbac_policy.go
```diff
@@ -152,3 +171,78 @@ func ParseRBACPoliciesField(buf []byte) ([]RBACPolicy, error) {
...
+func (a *AnonymousRBACPolicy) UnmarshalJSONFrom(dec *jsontext.Decoder) error {
+	var s serializedAnonymousRBACPolicy
+	err := json.UnmarshalDecode(dec, &s)
+	if err != nil {
+		return err
+	}
+
```
SuperSandro2000, line 228:
```suggestion
    if s.Permissions == "" {
      return nil
    }
```

internal/processor/accounts.go
```diff
@@ -158,19 +158,32 @@ func (p *Processor) CreateOrUpdateAccount(ctx context.Context, account keppel.Ac
...
+				anonPolicies = append(anonPolicies, p)
+			}
 		}
 		buf, _ := json.Marshal(account.RBACPolicies)
 		targetAccount.RBACPoliciesJSON = string(buf)
+
+		buf, err := json.Marshal(anonPolicies)
+		if err == nil && len(buf) < models.AnonymousRBACPoliciesJSONMaxLength {
```
SuperSandro2000, line 182:
```suggestion
		if err == nil && len(buf) <= models.AnonymousRBACPoliciesJSONMaxLength {
```

SuperSandro2000 (commented):
Not a super big fan AnonymousRBACPoliciesJSON = "" vs AnonymousRBACPoliciesJSON = []

internal/keppel/rbac_policy.go
majewsky: It is good practice to always do `*a = ...` in a `return nil` path to clearly spell out the expected state of the object post-unmarshaling

But then it would already be 5 extra lines for this case, when I don't see why we need a special branch for this. If `s.Permissions == ""`, then the next loop will not do anything and yield the same result.

majewsky:
> Not a super big fan AnonymousRBACPoliciesJSON = "" vs AnonymousRBACPoliciesJSON = []

I added a check constraint to enforce `"[]"` over `""`. Which then caused me to notice that the various `..._policies_json` fields in `accounts` are all over the place in this regard, so I also added check constraints for these. This expands the scope of the PR a bit, so if you feel it's too big to review, I can also split this part out into its own PR.

internal/processor/accounts.go
```diff
@@ -160,19 +160,32 @@ func (p *Processor) CreateOrUpdateAccount(ctx context.Context, account keppel.Ac
...
+
+			if policy, ok := anonPolicy.Unpack(); ok {
+				anonPolicies = append(anonPolicies, policy)
+			}
 		}
 		buf, _ := json.Marshal(account.RBACPolicies)
 		targetAccount.RBACPoliciesJSON = string(buf)
+
+		buf, err := json.Marshal(anonPolicies)
+		if err == nil && len(buf) <= models.AnonymousRBACPoliciesJSONMaxLength {
```
SuperSandro2000, line 184: Can we use some other estimation to end the loop early if it is likely to big? Right now we would build the entire string up to all to only throw it away if it is longer than 64 chars. Maybe we use an amount and say no more than 5 entries?

SuperSandro2000 (commented):
otherwise LGTM

internal/processor/accounts.go
```diff
@@ -160,19 +160,32 @@ func (p *Processor) CreateOrUpdateAccount(ctx context.Context, account keppel.Ac
...
 			account.RBACPolicies[idx] = policy
+
+			if policy, ok := anonPolicy.Unpack(); ok {
+				anonPolicies = append(anonPolicies, policy)
+			}
 		}
 		buf, _ := json.Marshal(account.RBACPolicies)
 		targetAccount.RBACPoliciesJSON = string(buf)
+
+		buf, err := json.Marshal(anonPolicies)
```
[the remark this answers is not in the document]
majewsky: Incorrect. encoding/json routes into encoding/json/v2 as of Go 1.27.

internal/processor/accounts.go
majewsky: This is during PutAccount, which is a rare operation, so I find this acceptable. (Also, we're handling all the RBAC policies anyway, so it's only causing more work proportional to the existing amount of work.)

The optimization target is to avoid loading lots of data in ReducedAccount when not necessary.

internal/api/keppel/accounts_test.go
```diff
@@ -291,6 +291,43 @@ func TestAccountsAPI(t *testing.T) {
...
+		"account": jsonmatch.Object{
+			"name":           "second",
+			"auth_tenant_id": "tenant1",
+			"metadata":       nil,
+			"rbac_policies":  newRBACPoliciesJSON,
+		},
+	})
+	// in the diff below, it is noteworthy that accounts.anon_rbac_policies_json remains at its previous value of "[]"
+	tr.DBChanges().AssertEqual(`
+		UPDATE accounts SET rbac_policies_json = '[{"match_repository":"verylongverylongverylongverylong","permissions":["anonymous_pull"]},{"match_repository":"evenlongerevenlongerevenlongerevenlonger","permissions":["anonymous_pull"]}]' WHERE name = 'second';
```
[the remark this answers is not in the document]
majewsky: yeah... whatever

majewsky:
I force-pushed to address the docstring notes from the last Copilot run. The TODO that it claimed to be missing is actually there, and I will take that in the next PR for this feature.

SuperSandro2000 (approved):

PR #791, opened 2026-08-28
Catch media type mismatches when decompressing images

SuperSandro2000:


internal/tasks/manifests_test.go
```diff
@@ -611,11 +611,11 @@ func TestCheckTrivySecurityStatus(t *testing.T) {
 			UPDATE blobs SET blocks_vuln_scanning = TRUE WHERE id = 9 AND account_name = 'test1' AND digest = '%[16]s';
 			UPDATE trivy_security_info SET vuln_status = 'Critical', next_check_at = %[9]d, checked_at = %[8]d, check_duration_secs = 0, has_enriched_report = TRUE WHERE repo_id = 1 AND digest = '%[1]s';
 			UPDATE trivy_security_info SET next_check_at = %[9]d, checked_at = %[8]d, check_duration_secs = 0 WHERE repo_id = 1 AND digest = '%[2]s';
-			UPDATE trivy_security_info SET vuln_status = 'Unsupported', message = 'vulnerability scanning is not supported for uncompressed image layers above %[11]g GiB', next_check_at = %[10]d WHERE repo_id = 1 AND digest = '%[6]s';
+			UPDATE trivy_security_info SET vuln_status = 'Unsupported', message = 'vulnerability scanning is not supported because either an image layer size is uncompressed above %[11]g GiB or we could not decompress it', next_check_at = %[10]d WHERE repo_id = 1 AND digest = '%[6]s';
```
majewsky: I do not like that the error messages are vague like that. If customers are uploading malformed images, we should tell them explicitly. Please change `blobs.blocks_vuln_scanning` into a string enum type, with the database containing e.g. `too-large` for oversized blobs or `cannot-decompress` for malformed compressed blobs.

majewsky (changes requested):

internal/keppel/database.go
```diff
@@ -87,7 +87,7 @@ var sqlMigrations = map[int64]string{
 			validation_error_message TEXT        NOT NULL DEFAULT '',
 			can_be_deleted_at        TIMESTAMPTZ DEFAULT NULL,
 			media_type               TEXT        NOT NULL DEFAULT '',
-			blocks_vuln_scanning     BOOLEAN     DEFAULT NULL,
+			blocks_vuln_scanning     TEXT        DEFAULT NULL,
```
majewsky: The test failure is probably caused by this spurious edit of a past transaction.

majewsky (changes requested):

internal/models/blob.go
```diff
@@ -39,7 +39,35 @@ type Blob struct {
...
-	BlocksVulnScanning     Option[bool]      `db:"blocks_vuln_scanning"`
+	// BlocksVulnScanning records the result of the blob's pre-scan check:
+	// None       = not checked yet
+	// Some(...)  = checked; the enum value describes whether (and why) the blob blocks vulnerability scanning.
+	BlocksVulnScanning Option[BlobVulnScanningBlock] `db:"blocks_vuln_scanning"`
+}
+
+// BlobVulnScanningBlock contains the reasons why a blob may block vulnerability scanning.
+// "none" indicates that the blob has been checked and does not block scanning.
+type BlobVulnScanningBlock string
```
majewsky: Naming nitpicks:
```suggestion
type BlobVulnScanningBlockReason string
```

internal/models/blob.go
```diff
@@ -39,7 +39,35 @@ type Blob struct {
...
+
+const (
+	// BlobVulnScanningBlockNone indicates that the blob has been checked and does NOT block vulnerability scanning.
+	BlobVulnScanningBlockNone BlobVulnScanningBlock = "none"
+	// BlobVulnScanningBlockTooLarge indicates that the blob's uncompressed size exceeds blobUncompressedSizeTooBigGiB.
+	BlobVulnScanningBlockTooLarge BlobVulnScanningBlock = "too-large"
+	// BlobVulnScanningBlockCannotDecompress indicates that the blob's compressed payload could not be decompressed.
+	BlobVulnScanningBlockCannotDecompress BlobVulnScanningBlock = "cannot-decompress"
```
majewsky:
```suggestion
	// BlobVulnScanningNotBlocked indicates that the blob has been checked and does NOT block vulnerability scanning.
	BlobVulnScanningNotBlocked BlobVulnScanningBlock = "none"
	// BlobVulnScanningBlockedBySize indicates that the blob's uncompressed size exceeds blobUncompressedSizeTooBigGiB.
	BlobVulnScanningBlockedBySize BlobVulnScanningBlock = "too-large"
	// BlobVulnScanningBlockedByDecompressError indicates that the blob's compressed payload could not be decompressed.
	BlobVulnScanningBlockedByDecompressError BlobVulnScanningBlock = "cannot-decompress"
```

internal/models/blob.go
```diff
@@ -39,7 +39,35 @@ type Blob struct {
...
+// Blocks returns true when vulnerability scanning is blocked.
+func (b BlobVulnScanningBlock) Blocks() bool {
+	switch b {
+	case BlobVulnScanningBlockTooLarge, BlobVulnScanningBlockCannotDecompress:
+		return true
+	case BlobVulnScanningBlockNone:
+		return false
+	default:
+		panic(fmt.Sprintf("unexpected BlobVulnScanningBlock value: %q", b))
+	}
```
majewsky: Usually it's a good idea to panic on unexpected values, but since BlockNone is predictably the only value that acts special here, it's fine to keep it simple:
```suggestion
	return b == BlobVulnScanningBlockNone
```

internal/tasks/manifests.go
```diff
@@ -855,27 +857,44 @@ func (j *Janitor) checkPreConditionsForTrivy(ctx context.Context, account models
...
 			if err != nil {
 				return false, layerBlobs, err
 			}
 		}
 
-		if blob.BlocksVulnScanning == Some(true) {
+		if reason, ok := blob.BlocksVulnScanning.Unpack(); ok && reason.Blocks() {
 			securityInfo.VulnerabilityStatus = models.UnsupportedVulnerabilityStatus
-			securityInfo.Message = fmt.Sprintf("vulnerability scanning is not supported for uncompressed image layers above %g GiB", blobUncompressedSizeTooBigGiB)
+			securityInfo.Message = fmt.Sprintf("vulnerability scanning is not supported because either an image layer size is uncompressed above %g GiB or we could not decompress it", blobUncompressedSizeTooBigGiB)
```
majewsky: This should now use the more specific `reason` value.

majewsky (changes requested):

majewsky (approved):

PR #792, opened 2026-08-28
Allow updating a replica account's platform filter

SuperSandro2000:


internal/processor/accounts.go
```diff
@@ -195,32 +195,33 @@ func (p *Processor) CreateOrUpdateAccount(ctx context.Context, account keppel.Ac
...
 	// validate platform filter
-	if !originalAccountExists {
-		switch replicationStrategy {
-		case keppel.NoReplicationStrategy:
-			if account.PlatformFilter != nil {
-				return models.Account{}, keppel.AsRegistryV2Error(errors.New(`platform filter is only allowed on replica accounts`)).WithStatus(http.StatusUnprocessableEntity)
-			}
-		case keppel.FromExternalOnFirstUseStrategy:
+	//
+	// On an existing account, an update that omits `platform_filter` reuses the existing filter, analogous to the replication policy.
```
majewsky: doublecheck

docs/api-spec.md
```diff
@@ -188,7 +188,7 @@ The following fields may be returned:
...
+| `accounts[].platform_filter` | list of objects or omitted | Only allowed for replica accounts. If not empty, when replicating an image list manifest (i.e. a multi-architecture image, or a single-architecture images with signatures or attestations attached), only submanifests matching one of the given platforms will be replicated as part of the image list manifest. Each entry must have the same format as the `manifests[].platform` field in the [OCI Image Index Specification](https://github.com/opencontainers/image-spec/blob/master/image-index.md). <br/> Submanifests not matching a platform filter may still be replicated when they are requested directly, but they will not be considered as part of the image list manifest (and thus e.g. not be protected from garbage collection). <br/> The platform filter may be changed on an existing replica account, but the effects of such a change (e.g. garbage collection of submanifests that no longer match the filter, or replication of submanifests that now match it) are applied asynchronously by background jobs and will take on the order of hours to fully settle. |
```
majewsky:
```suggestion
| `accounts[].platform_filter` | list of objects or omitted | Only allowed for replica accounts. If not empty, when replicating an image list manifest (i.e. a multi-architecture image, or a single-architecture images with signatures or attestations attached), only submanifests matching one of the given platforms will be replicated as part of the image list manifest. Each entry must have the same format as the `manifests[].platform` field in the [OCI Image Index Specification](https://github.com/opencontainers/image-spec/blob/master/image-index.md). <br/> Submanifests not matching a platform filter may still be replicated when they are requested directly, but they will not be considered as part of the image list manifest (and thus e.g. not be protected from garbage collection). <br/> On internal replica accounts, the platform filter must match the platform filter of the primary account. On external replica accounts, the platform filter may be changed after the account was created, but the effects of such a change (e.g. garbage collection of submanifests that no longer match the filter, or replication of submanifests that now match it) are applied asynchronously by background jobs and will take on the order of hours to fully settle. |
```

internal/processor/accounts.go
```diff
@@ -195,32 +195,33 @@ func (p *Processor) CreateOrUpdateAccount(ctx context.Context, account keppel.Ac
...
-			// either by specifying the same filter explicitly or omitting it
-			upstreamPlatformFilter, err := p.GetPlatformFilterFromPrimaryAccount(ctx, peer, targetAccount)
-			if err != nil {
-				return models.Account{}, keppel.AsRegistryV2Error(err).WithStatus(http.StatusInternalServerError)
-			}
+		}
+		// else: keep originalAccount.PlatformFilter which is already in targetAccount
+	case keppel.OnFirstUseStrategy:
+		// for internal replica accounts, the platform filter must match that of the primary account,
+		// either by specifying the same filter explicitly or omitting it
```
majewsky, line 214: This property will be broken if we have an external replica account that also has internal replicas (like for our many org-wide mirror accounts), and the external replica account has its platform filter updated.

Please add a new sync job on the account level that checks each internal replica account once an hour, calls `GetPlatformFilterFromPrimaryAccount` on the account, and updates the platform filter if necessary.

majewsky (changes requested):

internal/keppel/database.go
```diff
@@ -314,6 +314,10 @@ var sqlMigrations = map[int64]string{
 		ALTER TABLE pending_blobs
 			ADD COLUMN last_seen_at TIMESTAMPTZ NOT NULL DEFAULT NOW();
 	`,
+	56: `
+		ALTER TABLE accounts
+			ADD COLUMN next_platform_filter_sync_at TIMESTAMPTZ DEFAULT NULL;
```
majewsky, line 328: I find it unclean that we need to handle the `NULL` case for replica accounts. I suggest adding a constraint so that it can only be scheduled for internal replica accounts (and must be scheduled for those), similar to the constraint added in migration 53.

```suggestion
		ALTER TABLE accounts
			ADD COLUMN next_platform_filter_sync_at TIMESTAMPTZ DEFAULT NULL;
		UPDATE accounts SET next_platform_filter_sync_at = NOW() WHERE upstream_peer_hostname != '';
		ALTER TABLE accounts
			ADD CONSTRAINT platform_filter_sync_on_replicas CHECK ((upstream_peer_hostname = '') = (next_platform_filter_sync_at = NOW()));
```

Then update the Processor to set this field to an appropriate non-null value on new internal replica accounts.

internal/tasks/accounts.go
```diff
@@ -59,3 +63,70 @@ func (j *Janitor) announceAccountToFederation(ctx context.Context, account model
 	_, err = j.db.Exec(accountAnnouncementDoneQuery, account.Name, j.timeNow().Add(j.addJitter(1*time.Hour)))
 	return err
 }
+
+var accountPlatformFilterSyncSearchQuery = sqlext.SimplifyWhitespace(`
+	SELECT * FROM accounts
+		WHERE upstream_peer_hostname != ''
+		AND (next_platform_filter_sync_at IS NULL OR next_platform_filter_sync_at < $1)
+	-- accounts that have never been synced first, then sorted by last sync
+	ORDER BY next_platform_filter_sync_at IS NULL DESC, next_platform_filter_sync_at ASC
```
majewsky: If the previous suggestion is taken, this query can be simplified:
```suggestion
		AND next_platform_filter_sync_at < $1
	ORDER BY next_platform_filter_sync_at ASC
```

internal/tasks/accounts.go
```diff
@@ -59,3 +63,70 @@ func (j *Janitor) announceAccountToFederation(ctx context.Context, account model
...
+
+func (j *Janitor) syncAccountPlatformFilter(ctx context.Context, account models.Account, labels prometheus.Labels) error {
+	nextSyncAt := j.timeNow().Add(j.addJitter(1 * time.Hour))
+
+	peer, err := keppel.FindPeer(ctx, j.db, account.UpstreamPeerHostName)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			logg.Error("cannot sync platform filter for account %q: unknown peer %q", account.Name, account.UpstreamPeerHostName)
+			_, err2 := j.db.Exec(accountPlatformFilterSyncDoneQuery, account.Name, nextSyncAt)
+			return err2
```
majewsky: For consistency, I would like for the semantic error to be returned instead of logged. We have [errext.WithCleanup](https://pkg.go.dev/go.xyrillian.de/gg/errext#WithCleanup) to merge a semantic error with an optional cleanup error:
```suggestion
			err = fmt.Errorf("cannot sync platform filter for account %q: unknown peer %q", account.Name, account.UpstreamPeerHostName)
			_, err2 := j.db.Exec(accountPlatformFilterSyncDoneQuery, account.Name, nextSyncAt)
			return errext.WithCleanup(err, "next_platform_filter_sync_at update", err2)
```
Same for all other error-returning branches.

internal/tasks/accounts_test.go
```diff
@@ -53,6 +53,58 @@ func TestAnnounceAccountsToFederation(t *testing.T) {
...
+		j2, s2 := setupReplica(t, s1, "on_first_use")
+		ctx := t.Context()
+
+		s2.Clock.StepBy(1 * time.Hour)
+
+		syncJob := j2.AccountPlatformFilterSyncJob(s2.Registry)
+
+		assert.ErrEqual(t, syncJob.ProcessOne(s2.Ctx), nil)
+		account := must.ReturnT(keppel.FindAccount(ctx, s2.DB, "test1"))(t)
+		assert.Equal(t, len(account.PlatformFilter), 0)
```
majewsky: Should use the structured type to show more useful log output on error:
```suggestion
		assert.Equal(t, account.PlatformFilter, nil)
```

internal/tasks/accounts_test.go
```diff
@@ -53,6 +53,58 @@ func TestAnnounceAccountsToFederation(t *testing.T) {
...
+		primary1.PlatformFilter = newFilter
+		must.SucceedT(t, models.AccountStore.Update(ctx, s1.DB, primary1))
+
+		// do another full round of syncs
+		assert.ErrEqual(t, syncJob.ProcessOne(s2.Ctx), nil)
+		assert.ErrEqual(t, syncJob.ProcessOne(s2.Ctx), nil)
+		assert.ErrEqual(t, syncJob.ProcessOne(s2.Ctx), sql.ErrNoRows)
+
+		account = must.ReturnT(keppel.FindAccount(ctx, s2.DB, "test1"))(t)
+		assert.Equal(t, account.PlatformFilter.IsEqualTo(newFilter), true)
```
majewsky:
```suggestion
		assert.Equal(t, account.PlatformFilter, newFilter)
```

internal/tasks/accounts_test.go
```diff
@@ -53,6 +53,58 @@ func TestAnnounceAccountsToFederation(t *testing.T) {
...
+		account = must.ReturnT(keppel.FindAccount(ctx, s2.DB, "test2"))(t)
+		assert.Equal(t, len(account.PlatformFilter), 0)
```
majewsky: As before:
```suggestion
		assert.Equal(t, account.PlatformFilter, nil)
```

majewsky (changes requested):

majewsky (dismissed):
Good to go, save for the merge conflict.

majewsky (approved):

PR #804, opened 2026-09-17
rollup DB migrations

majewsky:
I verified this by running

```
rm -rf .testdb
go test -count 1 ./internal/client
.testdb/pg_dump.sh -d testrepoclientbasic --schema-only > $(date +%s).sql
```

before and after the change, and observing that the output is identical (except for the randomly generated tokens in the `\restrict/\unrestrict` guard commands). Feel free to check my work on this.

Also, this applies the same structure as in Limes, where the baseline is in an SQL file for proper syntax highlighting.


wagnerd3 (approved):
verified via `pg_dump`, I did not at all check the code itself.

